PR #12, opened 2015-07-14
List all services available to control

### LINKED ISSUE #6: Feature Request: list command to show all available services to control
Currently only shows the services that are running and available services to run.

gapple:
Resolves #6 

This change makes the assumption that the status of all of the services listed can be checked via the existence of plist files.  If services are only managed by `brew services` this should be a safe assumption.

I've removed the PID column because PID isn't available for all running services through `launchctl`, since it only returns services owned by the current user. Formula names may not match up to the command that's run (e.g. mysql => mysqld, php56 => php-fpm), so checking via `ps` would probably require parsing the plist file.

I'm not sure when the 'stale' and 'unknown' states would have occurred previously.
In order to detect 'unknown' again, any values in `ServicesCli.running` but not in `Formula.installed` would need to be added to the output; however `ServicesCli.running` will only have items for the current user.

### Before
    $ brew services list
    memcached  started      609 /Users/geoff/Library/LaunchAgents/homebrew.mxcl.memcached.plist
    mysql      started      619 /Users/geoff/Library/LaunchAgents/homebrew.mxcl.mysql.plist
    $ sudo brew services list
    dnsmasq    started       46 /Library/LaunchDaemons/homebrew.mxcl.dnsmasq.plist
### After
    $ brew services list
    Name      Status  User  Plist
    dnsmasq   started root  /Library/LaunchDaemons/homebrew.mxcl.dnsmasq.plist
    memcached started geoff /Users/geoff/Library/LaunchAgents/homebrew.mxcl.memcached.plist
    mysql     started geoff /Users/geoff/Library/LaunchAgents/homebrew.mxcl.mysql.plist
    php56     stopped

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -175,16 +175,41 @@ module ServicesCli
       true
     end
 
-    # List all running services with PID and status and path to plist file, if available
+    # List all available services with status, user, and path to plist file
     def list
-      opoo("No %s services controlled by `#{bin}` running..." % [root? ? 'root' : 'user-space']) and return if running.empty?
-      running.each do |label|
-        if svc = Service.from(label)
-          status = !svc.dest.file? ? "#{Tty.red}stale  " : "#{Tty.white}started"
-          puts "%-10.10s %s#{Tty.reset}  %7s %s" % [svc.name, status, svc.pid ? svc.pid.to_s : '-', svc.dest.file? ? svc.dest : label]
-        else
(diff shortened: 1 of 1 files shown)


xeron:
This is cool stuff.

About the name, I think it should reflect service name which could be used to stop/start/restart. So if it says `php56` I should be able to do `brew services start php56`.

kabel:
:+1:

PR #16, opened 2015-08-29
Handle no plist.

mikemcquaid:
If no plist is available throw an error (instead of writing an empty plist file).

CC @gapple @DomT4 @xu-cheng


DomT4:
Seems fine to me. Can't see any downsides.

PR #17, opened 2015-09-01
fix list

hmcfletch:
If there isn't a name that is longer than 4 characters, then line 208 threw a:

`Error: undefined method 'length' for nil:NilClass`

since `{ ... }[:user]` was `nil`

Checking for `nil` now


hmcfletch:
My name happens to be 'Les' so I hit this corner case.

gapple:
Looks like the issue is actually when no services are started yet, and therefore `formulae.max_by` returns Nil.  Will fix in a moment.

Issue #15, opened 2015-08-05
Stop all or start all option


r00takaspin:
Most of my services are databases or key-value storages, sometimes it is important to easy start all services (for example when I am starting work on project) or stop all services. It it possible out of box?

gapple:
That isn't currently implemented, but should be possible to add.

antstorm:
@r00takaspin it should now be possible, #28

mikemcquaid:
Yup, fixed, thanks!

PR #18, opened 2015-09-09
RFC: Add --all flag

kroehre:
First attempt at #15.

Feel free to close this PR if I'm way off here.
I think It would make sense to get some tests in before attempting to complete this, as it's a pretty big change.
I'm not sure how best to test this so I'd be grateful for some direction there.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -121,9 +121,6 @@ module ServicesCli
       extend(FileUtils)
     end
 
-    # Access current service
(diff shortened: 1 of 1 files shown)


mikemcquaid:
@kroehre I'm up for the refactoring you're doing but I think it'd be better to try and add just the `--all` flag for now and then do the refactoring in a future pull request (where you add tests and perhaps even a Travis CI job). How does that sound? Thanks for the PR!

kroehre:
@mikemcquaid agreed. I'll take another look and open another PR.

PR #26, opened 2015-11-10
Avoid failure exit mode if service already started

thomasleese:
This provides a convenient way of starting a service only if it wasn't already running and avoids using `restart`, parsing the output of `list`, or missing other important error messages when starting a service as part of an installation script.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -253,7 +253,10 @@ module ServicesCli
 
     # Start a service
     def start
-      odie "Service `#{service.name}` already started, use `#{bin} restart #{service.name}`" if service.loaded?
+      if service.loaded?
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Stupid question: what's the downside of just using `restart`?

thomasleese:
Some databases can take a good ten or twenty seconds to stop and this is unnecessary if they're already running. It's nothing major, just to improve speed.

mikemcquaid:
I think it might be nicer to fixup start so it behaves like this as default? Feels weird to add another option.

thomasleese:
Yeah, that sounds fine too, I just didn't want to break already existing behaviour. Perhaps just print a message saying that the service was already started and don't have a failure exit code instead then?

mikemcquaid:
Sounds good to me :+1:

mikemcquaid:
Thanks!

PR #27, opened 2015-11-29
Stop all services

antstorm:
Hey,

This is a really amazing plugin, been using it for some time now. Here's a small addition that I often use which allows to stop all the running services with a since command.

Also removed all the `<br>`s from the comments, which I think were put there by mistake.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -5,11 +5,11 @@
 #
 # ## SYNOPSIS
 #
-# [<sudo>] `brew services` `list`<br>
-# [<sudo>] `brew services` `restart` <formula><br>
-# [<sudo>] `brew services` `start` <formula> [<plist>]<br>
+# [<sudo>] `brew services` `list`
+# [<sudo>] `brew services` `restart` <formula>
+# [<sudo>] `brew services` `start` <formula> [<plist>]
 # [<sudo>] `brew services` `stop` <formula>|--all
-# [<sudo>] `brew services` `cleanup`<br>
+# [<sudo>] `brew services` `cleanup`
 #
 # ## DESCRIPTION
 #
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Might be nicer to make this an `--all` flag to `start`/`stop`/`restart`?

antstorm:
@mikemcquaid it sure sounds nice, although I can't think of any use cases for starting/restarting all services. Usually you'd restart a service when you change a config, so that's certainly one service at a time operation. As for the `start` — you'd probably wanna start a set of selected services for a particular project, not all available at once.

I can change the syntax to the `--all` flag for `stop` though.

mikemcquaid:
I can think of one: you want to start them again after stopping them all or you want to check they are all able to restart OK. Would be good to add the flag to all.

antstorm:
@mikemcquaid right, no problem, I'll add the flag to all the commands. Btw, I assume that `restart`would only restart the running services, not all available.

mikemcquaid:
@antstorm Personally I think it's more obvious for `stop`/`start`/`restart` to restart all services with `--all` rather than just running/stopped. What do you think?

antstorm:
@mikemcquaid yeah, makes sense to act on all the available services when `--all` is passed. Right, so the way this thing is structured is not very flexible when adding options, so I'm gonna rewrite some of it to support that. Meanwhile I've modified this PR to accept `--all` for the `stop` command only. This is now ready to be merged, while I'm working on the other commands.

mikemcquaid:
@antstorm I think I'd rather wait until all commands are ready, just for consistency. Thanks for your great work here!

antstorm:
@mikemcquaid I've created a new PR that add `--all` option for all mentioned commands — #28. Closing this PR.

PR #28, opened 2015-12-10
Act on all services

antstorm:
Hi,

This is an extended version of #27, allowing to act on `--all` services when running `start`, `stop` or `restart`.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -5,11 +5,11 @@
 #
 # ## SYNOPSIS
 #
-# [<sudo>] `brew services` `list`<br>
-# [<sudo>] `brew services` `restart` <formula><br>
-# [<sudo>] `brew services` `start` <formula> [<plist>]<br>
-# [<sudo>] `brew services` `stop` <formula><br>
-# [<sudo>] `brew services` `cleanup`<br>
+# [<sudo>] `brew services` `list`
+# [<sudo>] `brew services` `restart` <formula>
+# [<sudo>] `brew services` `start` <formula> [<plist>]
+# [<sudo>] `brew services` `stop` <formula>
+# [<sudo>] `brew services` `cleanup`
 #
 # ## DESCRIPTION
 #
@@ -83,6 +83,10 @@
 #     $ brew services list
 #     $ sudo brew services list
 #
+# Stop all running services for current user:
+#
+#     $ brew services stop --all
+#
 # ## BUGS
 #
 # `brew-services.rb` might not handle all edge cases, though it tries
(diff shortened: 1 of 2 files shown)


cmd/brew-services.rb
```diff
@@ -121,20 +125,38 @@ def homebrew!
...
 
-    # Access current service
-    def service; @service ||= Service.new(Formula.factory(@formula)) if @formula end
+    # Access target services
+    def target_services
+      @target_services ||= begin
+        if @formula == '--all'
+          available_services
+        elsif @formula
+          Array(single_service)
```
mikemcquaid: `[single_service]` is a bit more idiomatic, I think.

cmd/brew-services.rb
```diff
@@ -121,20 +125,38 @@ def homebrew!
...
+        end
+      end
+    end
+
+    # Access a single selected service
+    def single_service
+      @single_service ||= Service.new(Formula.factory(@formula)) unless @formula == '--all'
```
mikemcquaid: Maybe just `return nil if @formula == '--all'`?

cmd/brew-services.rb
```diff
@@ -171,36 +193,32 @@ def run!
 
     # Check if formula has been found
     def check
-      odie("Formula missing, please provide a formula name") unless service
+      odie("Formula(s) missing, please provide a formula name or use --all") unless target_services
```
mikemcquaid: `Formula(s)` to `Formula(e)`

mikemcquaid:
Looks pretty good to me but I want to do some more local testing. Thanks!

cmd/brew-services.rb
antstorm: Which is exactly the same result, do you think it's more readable?

cmd/brew-services.rb
antstorm: My bad, will fix that

antstorm:
@mikemcquaid sure, feel free to test it. I did some testing myself and couldn't find any issues, but you are probably more aware of edge cases. I'll fix the other comments meanwhile.

cmd/brew-services.rb
mikemcquaid: I think a little but up to you, I don't feel strongly.

cmd/brew-services.rb
```diff
@@ -121,20 +125,39 @@ def homebrew!
       extend(FileUtils)
     end
 
-    # Access current service
-    def service; @service ||= Service.new(Formula.factory(@formula)) if @formula end
+    # Access target services
+    def target_services
+      @target_services ||= begin
+        if @formula == '--all'
```
mikemcquaid: Was thinking it might be nice to split `@formula == '--all'` logic everywhere into a method like `def all_services?` and also perhaps make use of `ARGV.include?` and/or `ARGV.delete` instead.

antstorm: That's a good idea!

mikemcquaid:
Runs great locally, nice work. My only thought is it would be nice if `restart` restarted each service in order rather than stopping them all and then starting them all. Make sense?

antstorm:
@mikemcquaid Thanks! I've added the sequential restart. It's a bit hacky, but works and is self-container. I think I know the way to improve this even further, but that's something for the next time.

cmd/brew-services.rb
antstorm: Please take a look at the latest change and tell me if that's something that you had in mind :)

cmd/brew-services.rb
```diff
@@ -152,55 +179,52 @@ def run!
...
+      else
+        onoe "Unknown command `#{@cmd}`"
+        usage(1)
       end
     end
 
     # Check if formula has been found
     def check
-      odie("Formula missing, please provide a formula name") unless service
+      odie("Formula(e) missing, please provide a fo rmula name or use --all") unless target_services
```
mikemcquaid: `fo rmula` to `formula`

antstorm: Somehow missed this one, fixed now

antstorm:
@mikemcquaid I assume this one it good to go?

mikemcquaid:
Yeh, sorry, been busy and need to get time for final review/testing here, thanks!

mikemcquaid:

antstorm:
@mikemcquaid thanks mate, happy to be a part of it!

PR #29, opened 2015-12-22
Refactor out the use of global variables

antstorm:
Hey,

This PR shouldn't change the existing behaviour, it just reduces the number of global variables used by passing target directly into the method. Also makes the `restart` method much nicer.

Notice the use of `Array(x)` vs `[x]`, these do not have the same effect. `Array(x)` won't wrap a passed in array, while `[x]` will do that:

```ruby
irb(main):001:0> Array(1) == Array([1])
=> true
irb(main):002:0> [1] == [[1]]
=> false
```

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -125,27 +125,6 @@ module ServicesCli
       extend(FileUtils)
     end
 
-    # Access target services
-    def target_services
-      @target_services ||= begin
-        if act_on_all_services?
-          available_services
-        elsif @formula
-          [single_service]
-        end
-      end
-    end
-
-    # Access a single selected service
-    def single_service
-      return nil if act_on_all_services?
-      @single_service ||= Service.new(Formula.factory(@formula))
-    end
-
-    def act_on_all_services?
-      @act_on_all_services
-    end
-
     # All available services
     def available_services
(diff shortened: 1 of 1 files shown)


cmd/brew-services.rb
```diff
@@ -270,30 +255,21 @@ def cleanup
...
-      @act_on_all_services = false
-
-      services.each do |service|
-        @target_services = [service]
-        @single_service = service
-
-        stop if service.loaded?
-        start
+    def restart(target)
+      Array(target).each do |service|
```
mikemcquaid: Not disagreeing but what's the advantage over `target.to_a`? I've always wondered.

mikemcquaid:
Liking this change :+1:

cmd/brew-services.rb
antstorm: `.to_a` and `Array()` are a bit different. `.to_a` relies on the internal implementation of that method and if in fact missing by default. While `Array()` will try calling `.to_a` on the passed in object, but will simply wrap it in `[]` in case it's missing. I prefer to think of it as a `.to_a` with a fallback. And since `Service` doesn't have `.to_a` implemented, it does the trick.

mikemcquaid:

PR #33, opened 2016-02-08
Add hint on stop

capripot:
Provides a hint that the service is started as `root` or `user` when trying to stop it as `user` or `root` respectively.

Say `nginx` is starting on boot as `root`, when you do `brew services list`:
```
Name         Status  User Plist
nginx        started root /Library/LaunchDaemons/homebrew.mxcl.nginx.plist
```

When you was executing `brew services stop nginx`, that was the output:
```
Error: Service `nginx` not running, wanna start it? Try `brew services start nginx`
```

Which is confusing because `nginx` is actually running, but as `root`.

After applying this patch, you will get this message:
```
Error: Service `nginx` not started as `capripot`, wanna start it? Try `brew services start nginx`
Service `nginx` is started as `root` tho, wanna stop it? Try `sudo brew services stop nginx`
```

It also uses the new method in the `list` method and fix the wording in the original message from `running` to `started`.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -193,17 +193,17 @@ module ServicesCli
       formulae = available_services.map do |service|
         formula = {
           :name => service.formula.name,
-          :status => false,
+          :started => false,
           :user => nil,
           :plist => nil
         }
 
-        if (ServicesCli.boot_path + "#{service.label}.plist").exist?
-          formula[:status] = true
+        if service.started?(as: :root)
+          formula[:started] = true
           formula[:user] = "root"
           formula[:plist] = ServicesCli.boot_path + "#{service.label}.plist"
-        elsif (ServicesCli.user_path + "#{service.label}.plist").exist?
-          formula[:status] = true
+        elsif service.started?(as: :user)
+          formula[:started] = true
           formula[:user] = ServicesCli.user
           formula[:plist] = ServicesCli.user_path + "#{service.label}.plist"
         end
@@ -221,7 +221,7 @@ module ServicesCli
 
       puts "#{Tty.white}%-#{longest_name}.#{longest_name}s %-7.7s %-#{longest_user}.#{longest_user}s %s#{Tty.reset}" % ["Name", "Status", "User", "Plist"]
       formulae.each do |formula|
-        puts "%-#{longest_name}.#{longest_name}s %s %-#{longest_user}.#{longest_user}s %s" % [formula[:name], formula[:status] ? "#{Tty.green}started#{Tty.reset}" : "stopped", formula[:user], formula[:plist]]
+        puts "%-#{longest_name}.#{longest_name}s %s %-#{longest_user}.#{longest_user}s %s" % [formula[:name], formula[:started] ? "#{Tty.green}started#{Tty.reset}" : "stopped", formula[:user], formula[:plist]]
       end
     end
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Ideally I think we'd stop services being started as `root` unless they are `LaunchDaemons`.

capripot:
That makes sense, but what about a server started on port 80 for instance? If you need to restart it, you need to be root. Isn't it a use case?

mikemcquaid:
Yeh, in that case it should be a `LaunchDaemon` rather than a `LaunchAgent` and Homebrew will recommend it's installed into that location.

capripot:
Is `brew services [start|stop|restart]` made to manage `LaunchDaemon` as well? We could reword the error saying:
> Service  `nginx` is started as a `LaunchDaemon` so you need to be `root` to manage it, wanna stop it? Try `sudo brew services stop nginx`

cmd/brew-services.rb
```diff
@@ -304,7 +304,8 @@ def start(target, custom_plist = nil)
     def stop(target)
       if target.is_a?(Service) && !target.loaded?
         rm target.dest if target.dest.exist? # get rid of installed plist anyway, dude
-        odie "Service `#{target.name}` not running, wanna start it? Try `#{bin} start #{target.name}`"
+        odie "Service `#{target.name}` not started as `#{ServicesCli.root? ? "root" : ServicesCli.user}`, wanna start it? Try `#{bin} start #{target.name}`" +
+        ( target.started? ? "\nService `#{target.name}` is started as `#{target.started_as}` tho, wanna stop it? Try `#{ServicesCli.root? ? "" : "sudo "}#{bin} start #{target.name}`" : "" )
```
mikemcquaid: I think it would be nice to have this be a single error message rather than a new line with a separate message.

mikemcquaid:
@capripot Read the code a bit and decided I agree with your approach. Only tweak I'd like to see is a single error message rather than tacking on to the existing one. Thanks!

gapple:
The original "Service not running, wanna start it?" message seems weird to me, as an error when you're asking to stop a service.

The following error messages make sense to me:

    Error: Service `nginx` is not started.

    Error: Service `nginx` is started as `root`. Try `sudo brew services stop nginx`

    Error: Service `nginx` is started as `capripot`. Try `brew services stop nginx`

capripot:
It look way better and it addresses the one line error concern of @mikemcquaid, I think we can use these. Thanks!

cmd/brew-services.rb
```diff
@@ -304,7 +304,11 @@ def start(target, custom_plist = nil)
     def stop(target)
       if target.is_a?(Service) && !target.loaded?
         rm target.dest if target.dest.exist? # get rid of installed plist anyway, dude
-        odie "Service `#{target.name}` not running, wanna start it? Try `#{bin} start #{target.name}`"
+        if target.started?
+          odie "Service `#{target.name}` is started as `#{target.started_as}`. Try `#{ServicesCli.root? ? "" : "sudo "}#{bin} stop #{target.name}`"
```
mikemcquaid: `#{ServicesCli.root? ? "" : "sudo "}` could instead be `#{"sudo " if ServicesCli.root?}`

mikemcquaid:
One nit: otherwise really happy with this :+1:

cmd/brew-services.rb
capripot: Very true I also prefer this style Thanks

mikemcquaid:

capripot:
Thank you for managing the project, it's pretty useful :)
I just found out the message is reversed, the last commit fixes it.

PR #35, opened 2016-02-16
Hint message on stop is reversed

capripot:
Error message should show reverse, as user: `sudo..`, and as root: not sudo


gapple:
Thanks for the followup fix :+1:

Issue #38, opened 2016-03-23
Logic inverted in running services as root vs user


naturalethic:
When running as root, you need to add the UserName key to the plist, however, the key is being added when running as local user, and not when running as root, causing almost (if not all) services to fail.

naturalethic:
https://github.com/Homebrew/homebrew-services/pull/39

gapple:
I've copied your description to the PR, and will close this issue in favour or the pull request.

PR #39, opened 2016-03-23
Invert logic for adding UserName to plist when running as root or not

naturalethic:
When running as root, you need to add the UserName key to the plist, however, the key is being added when running as local user, and not when running as root, causing almost (if not all) services to fail.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -418,7 +418,7 @@ class Service
               gsub(%r{(<key>Label</key>\s*<string>)[^<]*(</string>)}, '\1' + label + '\2')
 
     # Force fix UserName
-    if !ServicesCli.root?
+    if ServicesCli.root?
       if data =~ %r{<key>UserName</key>}
         # Replace existing UserName value with current user
         data = data.gsub(%r{(<key>UserName</key>\s*<string>)[^<]*(</string>)}, '\1' + ServicesCli.user + '\2')
@@ -427,7 +427,7 @@ class Service
         data = data.gsub(%r{(\s*</dict>\s*</plist>)}, "\n    <key>UserName</key>\n    <string>" + ServicesCli.user + "</string>\\1")
       end
     elsif data =~ %r{<key>UserName</key>}
-      # Always remove UserName key entirely if running as root
+      # Always remove UserName key entirely if not running as root
       data = data.gsub(%r{(<key>UserName</key>\s*<string>)[^<]*(</string>)}, '')
     end
 


naturalethic:
https://github.com/Homebrew/homebrew-services/issues/38

gapple:
Can you provide more information on the particular formula that are failing for you?

I have not encountered any issues with the behaviour as it is currently.

naturalethic:
My tests were with mysql56, postgresql, and rabbitmq.  But just the logic is clear itself.  You can't launch a service with UserName specified if not running as root, right?

mikemcquaid:
Seems reasonable to me. @gapple @ilovezfs @xu-cheng any thoughts?

naturalethic:
Looks like the correct logic began to become confused here: https://github.com/Homebrew/homebrew-services/commit/5a8a136fcd91af1571bac356c0683eb28df7efc6

ilovezfs:
@mikemcquaid It's correct that UserName is only intended for use with root services in the system domain, but I'm failing to understand why we should ever set it at all.

mikemcquaid:
@ilovezfs I guess because that's the ownership of the Homebrew directory and running network services as root is :sob:

mikemcquaid:
Makes sense to me, thanks!

ilovezfs:
@mikemcquaid OK, but this is going to break the things that the incorrect commit was intended to fix.

ilovezfs:
If you run brew services as root, this is going to actively insert the UserName key and set it to a non-root user. That is contrary to what most launchd daemons expect to have happen. I understand why you'd want to replace the UserName in the UserName key with the correct user if that key is already there, but it makes no sense to actively insert it when it's not. If a particular brew service should not run as root but should be daemonized and activated via sudo brew services, then all the formula needs to do is specify the UserName key in the plist, and it will get corrected here with the right user. But if the UserName key is not there, the intent is a launchd root daemon with the user equal to root, which this commit now makes entirely impossible.

mikemcquaid:
@ilovezfs A fair point; have opened https://github.com/Homebrew/homebrew-services/pull/40 to revert until we talk more about strategy here.

> That is contrary to what most launchd daemons expect to have happen.

Do you think that still applies for most _Homebrew_ launchd daemons? The problem is I think we have two sets in a single bucket, currently:

1. LaunchAgents you want to run on machine startup regardless of login state
2. LaunchDaemons you want to be run on machine startup and always as root (either because they need root permissions to do something local, safe and rooty or because they will use them to e.g. bind to a priviledged port and then drop permissions).

ilovezfs:
>Do you think that still applies for most Homebrew launchd daemons?

Most launchd daemons as opposed to launchd user agents are written with the intent that they run as root.

Using the `UserName` key and/or the `GroupName` key is a corner case for the scenario you describe where something should be daemonized and always running even if no one is logged in. Despite the existence of the `UserName` key, I think most people would just put `sudo -u somenotrootdude` in their script as needed (or do the equivalent in whatever language).

I think requiring people to actually use the `UserName` key if they want a "non-root daemon" gives us the most flexibility without overturning the apple cart. Probably there should be some canonical brew user name that is inserted with the expectation that brew services will substitute in the username of whoever is running brew services. (Or longer term, an actual non-root user named `brew` (or whatever) set up when Homebrew, or at least homebrew services, is installed.) `vnstat.rb` uses `$USER`, which makes sense in the absence of a canonical name intended for substitution or an actual `brew` user.

Having formulae setting the `UserName` key to `root` is not at all idiomatic since a launchd daemon should be root anyway. But right now there are a bunch of formulae that do this, and this PR will replace "root" with a non-root user (`%x{/usr/bin/stat -f '%Su' #{HOMEBREW_CELLAR} 2>/dev/null}.chomp`), subverting those formulae's intent. I think those uses of the `UserName` key where they're setting `root` should probably just be removed altogether.

An alternative, which I would not recommend but would be an option, is to do this always-insert-or-replace-a-non-root-user thing *unless* the UserName key is present and the user in the formula is literally "root," but I find that bizarre.

mikemcquaid:
Good thoughts @ilovezfs. I think I'll personally leave this as-is for now but definitely open to more views here.

ilovezfs:
@mikemcquaid OK so that's going to voluntarily break at least the following:

bacula-fd
burp
dnscrypt-proxy
proftpd
shrewsoft-vpn-client
unbound
apache-archiva
twoping

mikemcquaid:
@ilovezfs When I say "leave this as-is for now" I mean "now that https://github.com/Homebrew/homebrew-services/pull/40 has been merged"

ilovezfs:
@mikemcquaid Ah, I didn't notice. :+1: 
My apologies.

I think the spirit of this PR is correct, which is that it never makes sense to set UserName for the launch agents, but I think Apple just ignores the key in that case, so it's mostly pedantry.

naturalethic:
I was getting errors in Console when UserAgent was set for LaunchAgents, so I think that part should remain.  Setting it on root wasn't correct though I see.  I was thinking of something like nginx which switches user, but that's after the service is run as root.  So derp on that.

mikemcquaid:
I think we can just remove setting the `User` for all `LaunchAgents`. That could be done both in `homebrew-services` and/or fixing plists in Homebrew itself. You up for picking either of those up @naturalethic? Sorry to create more work for you and thanks for this pull request either way.

naturalethic:
https://github.com/Homebrew/homebrew-services/pull/41

Any plist that set user to root (those mentioned above) should work after this patch.  Any plist that deliberately sets the UserName will be assumed to know what it's doing.  I think this is what we want?

gapple:
I regularly have both root (dnsmasq, httpd24) and non-root (memcached, mysql) services running without issue, and there haven't been any other reports of services failing, so I don't see how this was causing any failures.  I've additionally tried hard coding the username value for LaunchAgent to a non-existent user, and it still started successfully under the logged in user.

It's clear though that adding UserName to LaunchAgents has no effect, so that segment of code is useless and #41 seems like a safe change.

gapple:
The behaviour for LaunchDaemons was added due to [#14], but after this issue I think the correct behaviour is to leave plists as defined by default, and correct the formula.  The formula should know whether it needs to be launched as an Agent or Daemon, and the correct user as a Daemon (if necessary).  If a user needs to explicitly override the user when starting as a LaunchDaemon, it could be provided as an option.  This would potentially be a disruptive change for some users.

PR #41, opened 2016-03-28
Scrub username if root, otherwise leave it as is

naturalethic:


gapple:
As determined in #39 adding UserName to LaunchAgents has no effect, so that segment of code is useless and this looks like a non-disruptive change.

ilovezfs:
I think we need to just leave the code alone for now. This still does not look correct to me.

mikemcquaid:
@ilovezfs Could you elaborate?

ilovezfs:
I think this change is backwards. If we scrub the UserName key at all, which we probably shouldn't be doing in any case, it should be when it does *nothing*, which is when you *aren't* root, not when it does *something*, which is when you *are* root.

The change made in https://github.com/Homebrew/homebrew-services/pull/14 should be revisited with something that leaves the power in the hands of the formula to set a non-root user, rather than re-ratified as it is in this PR.

I think the question is how can we differentiate between
a) the UserName key contains a meaningful, specific non-root user that the daemon must run as
b) the UserName key contains a dummy `$USER` or <code>\`whoami\`</code> and expects to have that value replaced with the username of the administrator starting the service via `sudo brew services`
c) the UserName key contains `root` because the formula is trying to ensure it doesn't end up collateral damage from automagic key additions/removals/replacements.

*Removing the key when it does something, and not removing the key when it does nothing*, ends up just being an elaborate dance to ensure the key is always a no-op, which cannot possibly be what the formula writers are hoping for when they use the key ;)

PR #44, opened 2016-03-31
Change overriding plist UserName to require explicit option

gapple:
This replaces the current behaviour of always removing UserName from the plist when run as root with requiring a `--set-user` (or `-u`) option to explicitly override the user.

If `--set-user` is specified when not run as root, no change is applied and a warning is issued.

## Impact

This will affect plists that specify a non-root user and are run as LaunchDaemons. 
It should be up to the formula to determine if the UserName value should be specified in their plist by default.

e.g. 
 - <a href="https://github.com/Homebrew/homebrew-php/blob/8062c457835025d2f53e8096bed634ab5a723337/Abstract/abstract-php.rb#L578">PHP specifies the current user</a>
 - <a href="https://github.com/Homebrew/homebrew/blob/bf20a138788ae2355052619ab59a9b23c3b8e10a/Library/Formula/apache-archiva.rb#L41">apache-archiva specifies 'archiva'</a>
 - <a href="https://github.com/Homebrew/homebrew/blob/bf20a138788ae2355052619ab59a9b23c3b8e10a/Library/Formula/twoping.rb#L39">twoping specifies 'nobody'</a>

This change will have no effect on 
 - plists installed as LaunchAgents, since the UserName field is ignored and agents always run as the current user
 - plists installed as LaunchDaemons that do not include the UserName field
 - plists installed as LaunchDaemons that specify root as the UserName value

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -92,6 +92,9 @@
 # `brew-services.rb` might not handle all edge cases, but it will try to fix
 # problems if you run `brew services cleanup`.
 #
+
+require 'optparse'
+
 module ServicesCli
   class << self
     # Binary name.
@@ -133,21 +136,26 @@ module ServicesCli
     # Print usage and `exit(...)` with supplied exit code. If code
     # is set to `false`, then exit is ignored.
     def usage(code = 0)
-      puts "usage: [sudo] #{bin} [--help] <command> [<formula>|--all]"
-      puts
-      puts "Small wrapper around `launchctl` for supported formulae, commands available:"
-      puts "   cleanup Get rid of stale services and unused plists"
-      puts "   list    List all services managed by `#{bin}`"
-      puts "   restart Gracefully restart service(s)"
-      puts "   start   Start service(s)"
-      puts "   stop    Stop service(s)"
-      puts
-      puts "Options, sudo and paths:"
-      puts
-      puts "  sudo   When run as root, operates on #{boot_path} (run at boot!)"
-      puts "  Run at boot:  #{boot_path}"
-      puts "  Run at login: #{user_path}"
-      puts
+      puts <<-EOS.undent
+        Usage: [sudo] #{bin} [--help] [options] <command> [<formula>|--all]
+
+        Small wrapper around `launchctl` for supported formulae, commands available:
+          cleanup  Get rid of stale services and unused plists
+          list     List all services managed by `#{bin}`
+          restart  Gracefully restart service(s)
+          start    Start service(s)
+          stop     Stop service(s)
+
+          By default services are installed as LaunchAgents and run on login.
+          When `#{bin}` is run as root, services are installed as LaunchDaemons and run
+          on boot.
+
+        Options:
+
+          --set-user  When starting or restarting a service, modify the plist to run as
+                      the specified user.  This option is only available when run as
+                      root.
+      EOS
       exit(code) unless code == false
       true
     end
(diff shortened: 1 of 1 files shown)


ilovezfs:
@gapple Cool.

Some questions

1. Should we have formulae use `@user@` (or something else?) if they want the user replaced, and `some_specific_username` if they want it untouched in order to determine the default behavior?
2. Should the option be `--set-username` to avoid confusion?
3. Should `--set-username` take an optional argument if I don't want it set to me specifically (e.g., `--set-username=scapegoat`)?
4. How would you specify that you want the key removed? `--remove-username`? `--no-set-username`? Something else?

cmd/brew-services.rb
```diff
@@ -417,18 +440,19 @@ def generate_plist(data = nil)
...
 
-    # Force fix UserName
-    if !ServicesCli.root?
-      if data =~ %r{<key>UserName</key>}
-        # Replace existing UserName value with current user
-        data = data.gsub(%r{(<key>UserName</key>\s*<string>)[^<]*(</string>)}, '\1' + ServicesCli.user + '\2')
+    # Override plist UserName
+    if !set_user.nil?
+      if !ServicesCli.root?
+        opoo "--set-user has no effect for LaunchAgents"
```
ilovezfs: Should this be `odie` instead?

cmd/brew-services.rb
```diff
@@ -133,21 +136,26 @@ def available_services
...
+        Usage: [sudo] #{bin} [--help] [options] <command> [<formula>|--all]
+
+        Small wrapper around `launchctl` for supported formulae, commands available:
+          cleanup  Get rid of stale services and unused plists
+          list     List all services managed by `#{bin}`
+          restart  Gracefully restart service(s)
+          start    Start service(s)
+          stop     Stop service(s)
+
+          By default services are installed as a LaunchAgents and run on login.
```
ilovezfs: s/as a LaunchAgents/as LaunchAgents/

mikemcquaid:
Personally I'm 👎 on this option. The way `brew services` has worked is to take things like this from the formulae rather than requiring manual intervention. Use of this argument means that it could be easy to create subtly broken plists which are hard to debug.

What's the underlying user problem you're trying to deal with here?

ilovezfs:
@mikemcquaid As I said above,
>Should we have formulae use @user@ (or something else?) if they want the user replaced, and some_specific_username if they want it untouched in order to determine the default behavior

Would it make sense to handle it automagically based on a convention like that in the formulae rather than making it a command line option here?

```
    PHP specifies the current user
    apache-archiva specifies 'archiva'
    twoping specifies 'nobody'
```
Right now those formula values are effectively being replaced with root apparently contrary to what the formulae's authors wanted.

gapple:
1. If a formula wants to use the current user, they should use a similar approach to what PHP currently does.  The user shouldn't change between the time the formula is installed and when it's started with `brew services`, and imposing a convention such as `@user@` would potentially break the plist if it was run according to the instructions in `brew info <formula>`.
2. I don't feel strongly either way.  If `--set-username` better indicates that it's modifying the plist then :+1:
3. I think there would be ambiguity whether no argument would mean the current user or root (or maybe even to remove the value).  I would rather require the user be explicit than to make an assumption.
4. I don't think having a remove option is necessary.  It would be equivalent to `--set-username=root`.
5. That would probably be a helpful addition.  Using a non-existing user currently causes a sort of silent failure, where the plist is put in place but the desired program is not started.

--

This change restores control of behaviour to the plist, but enables users to override the behaviour if they know what they are doing.  The argument is not required, and in almost all cases should not be needed.

I did a grep that included homebrew, homebrew-apache, homebrew-dupes, homebrew-php, homebrew-science, and homebrew-versions. 
There were 11 formulae that specified UserName, only five of which specified a non-root user.

cmd/brew-services.rb
gapple: Other programs fail with an error for unknown options, so I'm not opposed to aligning with that convention.
I suppose if we assume that if someone is using the option they know that it needs sudo and just forgot, then an error will save them having to stop the service before running start again with sudo.

ilovezfs:
@gapple I'm also :-1: on this PR. After some discussion on Slack, I believe this is where we've landed:

1. Do away with virtually all current uses of the `UserName` key in all of the formulae's plists in core and every tap.
2. Change the code in `brew services` to never touch the `UserName` key at all, ever.
3. Allow possible exceptions to (1) on a case-by-case basis if the formula sets `UserName` key to something other than `root` or `$USER` or <code>\`whoami\`</code>, but that is not the concern of `brew services`, only the concern of code review of formulae.
4. Possibly provide a DSL method in brew itself for formulae to set `UserName` key to `nobody` in their plists, if there is enough demand for that. That, too, is not something `brew services` itself would have anything to do with or ever modify.

mikemcquaid:
@ilovezfs's plan sounds good to me too.

mikemcquaid:
@gapple Any news on this?

MikeMcQuaid:
Passing on this.

gapple:
I'm okay with passing on this, until there's a noted user need for such functionality.  I think in most cases, removing `UserName` from plists should be the correct approach.

PR #45, opened 2016-04-09
improve readme slightly

jaustinhughey:
Fixes typo in final example (`brew services start|stop|etc`) and adds an example to show users how to `brew services list` to confirm for anyone skimming that yes, said functionality is indeed there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ $ brew install dnsmasq
 $ sudo brew services start dnsmasq
 ```
 
-### What services might you have available? Let's find out! ###
+### List all services managed by `brew services` ###
 
 ```
 $ brew services list
--- a/README.md
+++ b/README.md
@@ -44,8 +44,21 @@ $ brew install dnsmasq
 $ sudo brew services start dnsmasq
 ```
 
+### What services might you have available? Let's find out! ###
+
+```
+$ brew services list
+Name       Status  User Plist
+mysql      stopped
+postgresql stopped
+rethinkdb  started xxx  /Users/xxx/Library/LaunchAgents/homebrew.mxcl.rethinkdb.plist
+```
+
+(This is only an example of output; your output will very likely deviate from the above.)
+
+
 ### Start/stop/restart all available services ###
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,8 +44,21 @@ $ brew install dnsmasq
 $ sudo brew services start dnsmasq
 ```
 
+### What services might you have available? Let's find out! ###
```
mikemcquaid: I believe rather than available this is what services have been started at some point with `brew services`.

README.md
```diff
@@ -44,8 +44,21 @@ $ brew install dnsmasq
...
+
+```
+$ brew services list
+Name       Status  User Plist
```
mikemcquaid: No other command usage shows output so I wonder if we should just omit it.

README.md
jaustinhughey: I'm 99.9% sure this *is* in fact what's available. Reason is, I didn't even know if `brew services list` existed until I straight up tried it to find out, and it listed all the "services" I'd installed in the past. For example, Cassandra, MySQL, PostgreSQL, rethinkdb. Never used it to start any of those before (since I just found out about this project).

README.md
jaustinhughey: In an ideal world they would all show output examples (provides point of reference for noobs if something goes wrong), but in this case I agree; I'll update the PR and nuke those output examples.

mikemcquaid: Thanks.

README.md
mikemcquaid: Maybe just `What services have you installed with 'brew services'? Let's find out!`?

gapple: The difference of tone stands out from other examples, which are all written in a more imperative mood.

The text from `brew services --help` is "``List all services managed by `brew services` ``"

jaustinhughey: Update incoming to match tone.

mikemcquaid:
Looks good! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #47, opened 2016-04-20
brew services cannot run under tmux.

mikemcquaid:
This is because `launchctl` gets upset about user attachments and very weird, hard-to-debug things happen. Note that we have a warning for this already in Homebrew. CC @xu-cheng

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -157,6 +157,8 @@ module ServicesCli
       homebrew!
       usage if ARGV.empty? || ARGV.include?('help') || ARGV.include?('--help') || ARGV.include?('-h')
 
+      odie "brew services cannot run under tmux!" if ENV["TMUX"]
+
       # Parse arguments.
       act_on_all_services = !!ARGV.delete('--all')
       args = ARGV.reject { |arg| arg[0] == 45 }.map { |arg| arg.include?("/") ? arg : arg.downcase } # 45.chr == '-'


xu-cheng:
> This is because launchctl gets upset about user attachments and very weird, hard-to-debug things happen.

But if you use reattach-to-user-namespace, it will work just fine. By introducing this change, you prevent users to use this tool under tmux completely.

xu-cheng:
c.f. https://github.com/ChrisJohnsen/tmux-MacOSX-pasteboard#mac-os-x-pasteboard-access-under-tmux-and-screen

mikemcquaid:
@xu-cheng Given this matches what we have in Homebrew now: can we please 🚢 this as-is? If you have a way of detecting whether `reattach-to-user-namespace` has been run I'm happy to relax that check but given the default case is broken I'd rather act on that.

mikemcquaid:
Merging this as-is for now but happy to make this error more helpful if we can detect the user namespace stuff.

xu-cheng:
I think you could make it just a warning instead of dying hard, which is what we real have in Homebrew.

ilovezfs:
You can check the exit code of `pbpaste`.

mikemcquaid:
@xu-cheng The difference to Homebrew is that we're not running the `launchctl` commands for you there but we are here.
@ilovezfs that feels like it could vary on e.g. the contents etc.

ilovezfs:
@mikemcquaid for pbpaste it should be 0 if you have it, and 1 if you don't. But if you want to check directly
```
(screen) Josephs-MacBook-Pro:~ joe$ tmux show-options -g default-command
default-command "reattach-to-user-namespace -l /bin/bash"
(screen) Josephs-MacBook-Pro:~ joe$ tmux show-options default-command
(screen) Josephs-MacBook-Pro:~ joe$ 
```
Also, right now the reattach-to-user-namespace formula has
```
  option "with-wrap-pbcopy-and-pbpaste", "Include wrappers for pbcopy/pbpaste that shim in this fix"
  option "with-wrap-launchctl", "Include wrapper for launchctl with this fix"
```
If it's useful for brew services to have the latter, should it be made recommended? Probably both options should be.

ilovezfs:
Hmm, I see those are command wrappers not functionality. I guess I wouldn't want those installed by default into my path, but it might be possible to have brew services install reattach-to-user-namespace and invoke launchctl through a wrapper like the optional one in the formula and not install the wrapper in the user's path since there'd be no reason to.

mikemcquaid:
> I guess I wouldn't want those installed by default into my path

In that case we probably need to keep this as-is. Ideally we'd just detect something finer grained than `ENV["TMUX"]`

ilovezfs:
brew services could invoke launchctl through a wrapper even if that wrapper isn't in my path.

mikemcquaid:
How does it know you have that wrapper installed or know where to look for it? I don't mind what we do as I think this change is fine as-is and it will help avoid people being confused when this fails (as I was helping someone debug why `brew services` wasn't working the other day).

ilovezfs:
I was imagining the wrapper being part of brew services not something installed separately to look for.

mikemcquaid:
Ok. Feels a bit overkill/out of scope considering people can just run it not in `tmux`.

mikemcquaid:
As generally a `brew services` invocation is a one-time thing anyway

ilovezfs:
I think the pbpaste thing would work since the only way for it to exit 0 from within tmux is if the namespace thing is set up.

ilovezfs:
```
>> system "pbpaste", "&>/dev/null"
system "pbpaste", "&>/dev/null"=> true
>> puts $?.exitstatus
0
=> nil
>>
```

```
>> system "pbpaste", "&>/dev/null"
=> false
>> puts $?.exitstatus
1
=> nil
>> 
```

ilovezfs:
This seems to work
```
odie "brew services cannot run under tmux!" if ENV["TMUX"] && !system("/usr/bin/pbpaste", "&>/dev/null")
```

mikemcquaid:
I'm dubious there's no other situation in which it would exit `1`? No empty clipboard etc.?

ilovezfs:
```
Josephs-MacBook-Pro:~ joe$ echo foo | pbcopy 
Josephs-MacBook-Pro:~ joe$ pbpaste 
foo
Josephs-MacBook-Pro:~ joe$ pbcopy < /dev/null
Josephs-MacBook-Pro:~ joe$ pbpaste 
Josephs-MacBook-Pro:~ joe$ echo $?
0
```

mikemcquaid:
Ok, seems reasonable. Mind making the PR for here and Homebrew/brew? 🙇

ilovezfs:
Sure, no problem.

ilovezfs:
@mikemcquaid 
https://github.com/Homebrew/brew/pull/130
https://github.com/Homebrew/homebrew-services/pull/48

PR #48, opened 2016-04-21
Allow tmux if we detect reattach-to-user-namespace

ilovezfs:
The exit status of pbpaste is a proxy for whether the user has set up
reattach-to-user-namespace. It should be 0 if it is set up, 1 otherwise.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -173,7 +173,10 @@ module ServicesCli
       homebrew!
       usage if ARGV.empty? || ARGV.include?("help") || ARGV.include?("--help") || ARGV.include?("-h")
 
-      odie "brew services cannot run under tmux!" if ENV["TMUX"]
+      # pbpaste's exit status is a proxy for detecting the use of reattach-to-user-namespace
+      if ENV["TMUX"] && !quiet_system("/usr/bin/pbpaste")
+        odie "brew services cannot run under tmux!"
+      end
 
       # Parse arguments.
       act_on_all_services = !ARGV.delete("--all").nil?


cmd/brew-services.rb
```diff
@@ -173,7 +173,8 @@ def run!
       homebrew!
       usage if ARGV.empty? || ARGV.include?("help") || ARGV.include?("--help") || ARGV.include?("-h")
 
-      odie "brew services cannot run under tmux!" if ENV["TMUX"]
+      # pbpaste's exit status is a proxy for detecting the use of reattach-to-user-namespace
+      odie "brew services cannot run under tmux!" if ENV["TMUX"] && !system("/usr/bin/pbpaste", "&>/dev/null")
```
mikemcquaid: Maybe use the block form of `if` to make this more readable/less long a line. Also, using `"&>/dev/null"` as a separate argument won't do what you think it will; it'll pass that as an argument. Just remove embed the space in there as there's no user-submitted input we're worried about here.

ilovezfs:
@molovo does this resolve #49 for you?

molovo:
Practically, yes it does. No reason why this shouldn't be merged.

Don't know how others feel about this, but would it be worth showing a more informative error message for those who might not know that `reattach-to-user-namespace` exists? Just a `Please run: brew install reattach-to-user-namespace`. That's all that's needed as the `reattach-to-user-namespace` Formula contains and sets up all the necessary wrapper functions automatically.

ilovezfs:
@molovo I believe you still have to set `set-option -g default-command "reattach-to-user-namespace -l zsh"` unless you install the wrappers, which aren't the default and aren't in the bottles since they're only `with` options (optional) not `without` options (recommended/default):
https://github.com/Homebrew/homebrew-core/blob/master/Formula/reattach-to-user-namespace.rb#L17-L18

molovo:
Fair enough. Then maybe the message could say `brew install reattach-to-user-namespace --with-wrap-launchctl --with-wrap-pbcopy-and-pbpaste`. That way it's blatantly obvious to a user what they need to do to get round the error.

mikemcquaid:
I think I'd personally rather leave it to `tmux` users to figure that stuff out; I don't want to have to support the edge cases that may result. 👍 to 🚢 this.

PR #55, opened 2016-05-25
Cleanup argument handling, remove shebang.

MikeMcQuaid:
The shebang is no longer needed for this to be called by Homebrew and the argument handling is buggy and can be simplified by using `ARGV.named`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os: osx
 
 # Empty (to avoid Bundler).
 install:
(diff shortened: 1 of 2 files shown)


apjanke:
New version is way more readable to me and the behavior seems right, and the new `--all` semantics make more sense. Can't comment on possible regressions because I can't read the old version well enough to understand exactly what it's doing.

:+1:

PR #57, opened 2016-05-29
brew-services: handle invalid keg.

### LINKED ISSUE #54: Error: / is not inside a keg
I am running homebrew (Homebrew 0.9.9, git revision 898cb; last commit 2016-05-25) on 10.11.5 (15F34).

Starting yesterday the call to `brew services list` throws an error:

~~~bash
Error: / is not inside a keg
~~~

I removed all homebrew-related *.plist files and retried, but the error persists.

Do you have any insight or idea what this error points to?

MikeMcQuaid:
Closes #54.


UniqMartin:
:+1:

PR #58, opened 2016-05-30
Always remove UserName key as a stopgap

### LINKED ISSUE #56: Unable to start services - "The UserName key is not supported for non-System services."
Since upgrading to OS X 10.11.5 I can no longer start services like httpd22.

`brew services start httpd22` results in `==> Successfully started ``httpd22`` (label: homebrew.mxcl.httpd22)` which is perfectly fine.

However, when inspecting the logs using `syslog -k Sender com.apple.xpc.launchd` I noticed the following entries:
```bash
May 27 09:48:22 name com.apple.xpc.launchd[1] (homebrew.mxcl.httpd22) <Error>: The UserName key is not supported for non-System services.
May 27 09:48:22 name com.apple.xpc.launchd[1] (homebrew.mxcl.httpd22[22553]) <Warning>: Service exited with abnormal code: 1
```

The `.plist`-file contains the section in question which was perfectly fine prior the update. Anyone experienced anything like that before?

ilovezfs:
This will avoid setting the key inappropriately, but will also continue
to prevent "legitimate" use of the key (e.g., setting UserName "php" for
a daemon).

Closes #56.


ilovezfs:
I will defer to #59 :)

PR #59, opened 2016-05-30
Fixed error when starting service using OS X 10.11.5

### LINKED ISSUE #56: Unable to start services - "The UserName key is not supported for non-System services."

daFish:
This applies the fix from #56 

Fixes #56

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -466,17 +466,7 @@ class Service
     # Replace "template" variables and ensure label is always, always homebrew.mxcl.<formula>
     data = data.to_s.gsub(/\{\{([a-z][a-z0-9_]*)\}\}/i) { |_m| formula.send($1).to_s if formula.respond_to?($1) }.gsub(%r{(<key>Label</key>\s*<string>)[^<]*(</string>)}, '\1' + label + '\2')
 
-    # Force fix UserName
-    if !ServicesCli.root?
-      if data =~ %r{<key>UserName</key>}
-        # Replace existing UserName value with current user
-        data = data.gsub(%r{(<key>UserName</key>\s*<string>)[^<]*(</string>)}, '\1' + ServicesCli.user + '\2')
-      else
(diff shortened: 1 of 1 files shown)


cmd/brew-services.rb
```diff
@@ -466,17 +466,7 @@ def generate_plist(data = nil)
...
-      if data =~ %r{<key>UserName</key>}
-        # Replace existing UserName value with current user
-        data = data.gsub(%r{(<key>UserName</key>\s*<string>)[^<]*(</string>)}, '\1' + ServicesCli.user + '\2')
-      else
-        # Add UserName key and value to end of plist if it doesn't already exist
-        data = data.gsub(%r{(\s*</dict>\s*</plist>)}, "\n    <key>UserName</key>\n    <string>" + ServicesCli.user + "</string>\\1")
-      end
-    elsif data =~ %r{<key>UserName</key>}
-      # Always remove UserName key entirely if running as root
+    if data =~ %r{<key>UserName</key>}
```
MikeMcQuaid: Tiny tweak: a little comment saying something about always removing the UserName as it doesn't work on 10.11.5 would be 👍

MikeMcQuaid:
Thanks @daFish!

PR #69, opened 2016-09-07
Improve documentation of commands

### LINKED ISSUE #68: help output implies functionality which is not present
In particular;
```
   start   Start service(s)
   stop    Stop service(s)
```
seem to imply you can supply them _a list_ of services, which is not the case, the only plural case is `--all`, but that is only really made clear in the [more `man`-style help text in the source code](https://github.com/Homebrew/homebrew-services/blob/573611593c2d9132c190d77f5f3d14d5f54cc7d5/cmd/brew-services.rb#L8).

It seems like either the start and stop commands should accept a list (probably unlikely, given the "custom plist" functionality already present there!) or the help text should not imply plurality, or include the more helpful command signature from the header of the file.

I'd be happy to supply a PR if desired!

ticky:
- clarify that `restart`, `start` and `stop` will restart one named service, unless used with `--all`

fixes #68

💜

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -152,12 +152,13 @@ module ServicesCli
       puts "Small wrapper around `launchctl` for supported formulae, commands available:"
       puts "   cleanup Get rid of stale services and unused plists"
       puts "   list    List all services managed by `#{bin}`"
-      puts "   restart Gracefully restart service(s)"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock! 👏

PR #73, opened 2016-09-30
Add zsh completions for brew-services

joshka:
This addresses code that was lost from the changes to the completions
scripts from oh-my-zsh. It requires a change to brew to install
completion scripts, but can be manually tested before then via symlink:

```shell
ln -s /usr/local/Homebrew/Library/Taps/homebrew/homebrew-services/completions/zsh/_brew_services /usr/local/share/zsh/site-functions/_brew_services
```

### DIFF
--- a/completions/zsh/_brew_services
+++ b/completions/zsh/_brew_services
@@ -1,0 +1,103 @@
+#compdef brew
+#autoload
+
+__brew_installed_services() {
+  local -a services
+  services=($(brew services list | awk '{print $1}' | tail -n+2))
+  _describe -t services 'installed services' services
+}
+
+__brew_services_commands() {
+  local -a commands
+  commands=(
+    'cleanup:Get rid of stale services and unused plists'
+    'list:List all services managed by brew services'
+    'restart:Gracefully restart selected service'
+    'start:Start selected service'
+    'stop:Stop selected service'
+  )
+  _describe -t commands 'commands' commands
+}
+
+__brew_services_expand_alias()
+{
+  local command_or_alias="$1"
+  local -A aliases
+  aliases=(
+    clean cleanup
+    cl cleanup
+    rm cleanup
+    ls list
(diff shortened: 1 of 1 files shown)


joshka:
Fixes #73

MikeMcQuaid:
@joshka Somewhat relatedly: you may want to pull this into the code repo too now: https://github.com/Homebrew/homebrew-completions/blob/master/brew-cask-completion.rb

MikeMcQuaid:
> It requires a change to brew to install
> completion scripts, but can be manually tested before then via symlink:

👍 but will wait on the Homebrew/brew change before we merge this.

completions/zsh/_brew_services
```diff
@@ -0,0 +1,104 @@
+#compdef brew
+#autoload
+
+__brew_installed_services() {
+  local -a services
+  # services=($(brew services list | awk '{print $1}' | tail -n+2))
+  services=(a b c)
```
apjanke: Looks like some debugging code that was left in? I'm getting odd completion behavior that looks like it's from this.

```
[~]
$ brew services list
Name  Status  User Plist
mysql stopped
[~]
$ brew services start
a  b  c
```

joshka: Whoops - yeah I was testing without anything I wanted to install as a service ;)
Thanks for picking that up

joshka: Fixed in latest push

joshka:
@MikeMcQuaid I've asked the original author to re-license the brew-cask completions from MIT to BSD, and will get them into brew when that issue is resolved (either with or without an extra header with license info).
https://github.com/xyb/homebrew-cask-completion/issues/4

MikeMcQuaid:
@joshka Thanks!

MikeMcQuaid:
Thanks again @joshka!

PR #77, opened 2016-10-17
Prevent error message when cmd is nil

reacuna:


MikeMcQuaid:
In what situation is this `nil`? Thanks!

reacuna:
When executing `brew services` without a command

reacuna:
For example:
```
[system] ~ brew services
Error: Unknown command ``!
brew services
    Easily start and stop formulae via launchctl

    Integrates Homebrew formulae with OS X's launchctl manager. Services can be
(etc...)
```

MikeMcQuaid:
Nice catch, thanks! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

Issue #84, opened 2016-11-22
No errors for services that aren't installed


ekmartin:
If I'm not completely off it seems like all the commands complete successfully even if the target service isn't installed, as long as it is a valid brew service.

Example (I don't have RabbitMQ installed):
![image](https://cloud.githubusercontent.com/assets/3536982/20546581/18bace0e-b0e4-11e6-8633-1745bc8f4769.png)

`brew services start` successfully creates the `plist`, but it doesn't seem like it checks that the target program actually starts.

MikeMcQuaid:
As we just call `launchctl` we can't detect or handle if the application was actually started just that the start attempt was completed.

MikeMcQuaid:
Ah, I see you mean if it's not installed via Homebrew rather than the plist not being installed.

PR #85, opened 2016-11-23
Make clear start is requested not necessarily done

MikeMcQuaid:
We're not `launchctl` so we don't know if it's actually started or not (or started and died immediately) so set expectations accordingly.

CC @ilovezfs for thoughts.


MikeMcQuaid:
Passing on this after misunderstanding #84.

PR #87, opened 2016-11-24
Separate install and start

loderunner:
This is a proposal for separation of concerns. Sometimes, I want to start/stop a service without installing it to launch at login or boot. I tried to achieve this goal by making 2 new commands: `install` and `uninstall`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ brew tap homebrew/services
 
 ## Examples ##
 
-### Install and start service mysql at login ###
+### Start service mysql ###
 
 ```
 $ brew install mysql
@@ -36,12 +36,17 @@ Restart service mysql:
 $ brew services restart mysql
 ```
 
+Install and start service mysql at login:
+
+```
+$ brew services install mysql
+```
 
 ### Install and start dnsmasq service at boot ###
 
 ```
 $ brew install dnsmasq
-$ sudo brew services start dnsmasq
+$ sudo brew services install dnsmasq
 ```
 
 ### List all services managed by `brew services` ###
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
> This branch has conflicts that must be resolved

Squash commits by running `brew update; git checkout $YOUR_BRANCH; git rebase --interactive origin/master` and then mark each commit other than the first as `squash` or `fixup` then `git push -f`. Thanks!



> Sometimes, I want to start/stop a service without installing it to launch at login or boot. I tried to achieve this goal by making 2 new commands: install and uninstall.

That's fine but you need to do so without changing the behaviour of existing commands.

mareksuscak:
I was thinking the same lately. I often don't let services auto-start at boot because every single project I worked on required various services and I don't want them all started automatically and slow down the machine boot and consume energy when running on battery. 

An ideal solution would be the one @loderunner had implemented but I understand that you'd prefer not to break the existing behavior so how about adding a `--no-install` switch to the `start` command for now and add a warning it'll change in a few months time to let the community catch up? 

I'd keep `install` command though and at some point in the future make the `start` command do what it implies based on the name. I feel like it currently has two separate responsibilities and I was too confused by this fact when I initially used the command.

MikeMcQuaid:
> I don't want them all started automatically and slow down the machine boot and consume energy when running on battery.

FYI, they don't slow down boot as they are run in the background after you've logged in (or after the machine has already booted for startup plists) and they don't consume energy unless they are using CPU (i.e. you're actually hitting them).



> An ideal solution would be the one @loderunner had implemented but I understand that you'd prefer not to break the existing behavior so how about adding a --no-install switch to the start command for now and add a warning it'll change in a few months time to let the community catch up?

I don't see a good reason to break backwards compatibility here. As mentioned, I'm happy to add a new command, though.

MikeMcQuaid:
Implemented in https://github.com/Homebrew/homebrew-services/issues/96.

PR #90, opened 2016-12-08
Detect service error

antstorm:
Hi,

I've stumbled across a case when `postgresql` would report being started successfully, but won't have a running process. It turned out that `launchctl` wasn't receiving the error code on `load`, but `list` revealed a non-zero exit code.

With this change in place it will mark services that aren't actually running, although they are loaded.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -142,8 +142,12 @@ module ServicesCli
         end
 
         # Check the exit code of the service, might indicate an error
-        if formula[:status] == :started && service.exit_code.nonzero?
-          formula[:status] = :error
+        if formula[:status] == :started
+          if service.unknown_status?
+            formula[:status] = :unknown
+          elsif service.error?
+            formula[:status] = :error
+          end
         end
 
         formula
@@ -165,6 +169,9 @@ module ServicesCli
             "#{Tty.green}started#{Tty.reset}"
           when :stopped
             "stopped"
+          when :unknown
+            # For consistency showing unknown state as started in yellow colour
+            "#{Tty.yellow}started#{Tty.reset}"
           when :error
             "#{Tty.red}error  #{Tty.reset}"
           end
@@ -381,6 +388,14 @@ class Service
     nil
   end
 
+  def error?
+    !pid || pid.zero? || !exit_code || exit_code.nonzero?
+  end
+
+  def unknown_status?
+    !status || status.empty?
+  end
+
   # Get current PID of daemon process from launchctl.
   def pid
     return $1.to_i if status =~ status_regexp
@@ -423,11 +438,11 @@ class Service
   private
 
   def status
-    `#{ServicesCli.launchctl} list | grep #{label} 2>/dev/null`.chomp
+    @status ||= `#{ServicesCli.launchctl} list | grep #{label} 2>/dev/null`.chomp
   end
 
   def status_regexp
-    /\A([\d\-]+)\s+([\d]+)\s+#{label}\z/
+    /\A([\d-]+)\s+([\d]+)\s+#{label}\z/
   end
 end
 
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -126,19 +126,24 @@ module ServicesCli
       formulae = available_services.map do |service|
         formula = {
           name: service.formula.name,
-          started: false,
+          status: :stopped,
           user: nil,
           plist: nil,
         }
 
         if service.started?(as: :root)
-          formula[:started] = true
+          formula[:status] = :started
           formula[:user] = "root"
           formula[:plist] = ServicesCli.boot_path + "#{service.label}.plist"
         elsif service.started?(as: :user)
-          formula[:started] = true
+          formula[:status] = :started
           formula[:user] = ServicesCli.user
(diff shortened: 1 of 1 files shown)


cmd/brew-services.rb
```diff
@@ -155,9 +160,18 @@ def list
       puts format("#{Tty.white}%-#{longest_name}.#{longest_name}s %-7.7s %-#{longest_user}.#{longest_user}s %s#{Tty.reset}",
                   "Name", "Status", "User", "Plist")
       formulae.each do |formula|
+        status = case formula[:status]
+                 when :started
```
MikeMcQuaid: Indent `when :started` two-chars after `status`. Sorry for pedantry, a pet hate 😉

cmd/brew-services.rb
```diff
@@ -155,9 +160,18 @@ def list
...
+                   "#{Tty.green}started#{Tty.reset}"
+                 when :stopped
+                   "stopped"
+                 when :error
+                   "#{Tty.red}error  #{Tty.reset}"
```
MikeMcQuaid: Why the two spaces here? For alignment?

antstorm: Yes, exactly! It felt like a proper alignment of the output is a bit out of scope for this PR :)

cmd/brew-services.rb
antstorm: Happy to do that if that's what you prefer, I'm just following the community style guide here (https://github.com/bbatsov/ruby-style-guide#indent-conditional-assignment)

MikeMcQuaid: Yeh, I normally agree with that but this is one of the few cases I hate; it feels like it's floating in mid-air 😆. Thanks!

cmd/brew-services.rb
MikeMcQuaid: Yeh, definitely. Good to explain.

cmd/brew-services.rb
antstorm: No worries, mate, pushed that change :)

cmd/brew-services.rb
```diff
@@ -126,21 +126,26 @@ def list
...
           formula[:plist] = ServicesCli.boot_path + "#{service.label}.plist"
         elsif service.started?(as: :user)
-          formula[:started] = true
+          formula[:status] = :started
           formula[:user] = ServicesCli.user
           formula[:plist] = ServicesCli.user_path + "#{service.label}.plist"
         end
 
+        # Check the exit code of the service, might indicate an error
+        if formula[:status] == :started && service.exit_code.nonzero?
```
MikeMcQuaid: `service.exit_code` can be `nil` so that probably needs guarded.

cmd/brew-services.rb
```diff
@@ -369,8 +383,11 @@ def started_as
 
   # Get current PID of daemon process from launchctl.
   def pid
-    status = `#{ServicesCli.launchctl} list | grep #{label} 2>/dev/null`.chomp
-    return $1.to_i if status =~ /\A([\d]+)\s+.+#{label}\z/
+    return $1.to_i if status =~ status_regexp
+  end
+
+  def exit_code
+    return $2.to_i if status =~ status_regexp
```
MikeMcQuaid: An edge-case I noticed with local testing: if I `brew services list` then any services running as `root` will show up as `error` because they aren't listed in the non-root `brew services list` output. Not sure how best to handle that, though.

antstorm:
@MikeMcQuaid I've changed the solution a bit to support multiple users. So now it will report service status as "unknown" if we're unable to detect it's status via `launchctl list`. I've also added `pid` check, which is also an indicator that there was an error.

Here's an example running as regular user:



and as root:



P.S. Error next to `mysql` is due to a missing `pid`, which is expected.

MikeMcQuaid:
@antstorm Great work. Although the `unknown` is technically correct I wonder if we can use less scary language. Thoughts on just using `started` if it was started as a different user? Alternatively, making it something like `launched` or something?

antstorm:
@MikeMcQuaid sure, we can use started, but mark it yellow. Launched also makes sense, but might be a bit confusing since it's a synonym to started. Another option is `started`/`running`.

MikeMcQuaid:
I think communicating purely with colours isn't great for usability. Agreed about "launched" problems. Let's maybe just keep it as `started` for now; you're adding detection for a failure case and some people parse the output of this command so I'd probably rather have false-positives than false-negatives here as it still provides some improvement while not requiring anything that parses the output to be changed.

antstorm:
@MikeMcQuaid so keep things as they were when showing status of services started by other users?

MikeMcQuaid:
Yeh but just for when it would have been "unknown". Basically, rather than showing "unknown" we just fall back to started/stopped like the behaviour before this PR. Progressive enhancement rather than degradation, essentially.

antstorm:
@MikeMcQuaid pushed up that change

MikeMcQuaid:
Thanks for your truly fantastic contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

antstorm:
@MikeMcQuaid thanks, happy to help! :)

PR #95, opened 2017-01-09
Make a missing PID "unknown" rather than "error".

MikeMcQuaid:
This likely indicates a badly configured Homebrew `plist` but it's something that should be exposed to Homebrew contributors rather than users who find it confusing. CC @antstorm for thoughts and code review.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -165,16 +165,17 @@ module ServicesCli
                   "Name", "Status", "User", "Plist")
       formulae.each do |formula|
         status = case formula[:status]
-          when :started
-            "#{Tty.green}started#{Tty.reset}"
-          when :stopped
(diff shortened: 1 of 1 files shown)


antstorm:
Maybe it's worth adding a CLI option that would act similar to `HOMEBREW_DEVELOPER` flag in this context (since I'm not sure of the other side effects of setting `HOMEBREW_DEVELOPER` on the system). Otherwise looking good!

MikeMcQuaid:
@antstorm Any suggestions for a name?

antstorm:
`--debug / -d`… or maybe `--verbose / -v`, since it actually gives a bit more verbose output

MikeMcQuaid:
Let's leave it as-is for now but 👍 to having one of those flags output more debugging information.

PR #97, opened 2017-01-18
Fix run --all command.

mareksuscak:
I’ve found small bug in the newly implemented run command. This PR fixes it. Sorry for that.


MikeMcQuaid:
Thanks again @mareksuscak!

PR #103, opened 2017-02-04
Fix completion parse error

artdevjs:
This PR fixes the error ``_brew_services:88: parse error near `>'`` due unescaped single quote.
An alternative way is to use UTF apostrophe `’`.

### DIFF
--- a/completions/zsh/_brew_services
+++ b/completions/zsh/_brew_services
@@ -13,7 +13,7 @@ __brew_services_commands() {
     'cleanup:Get rid of stale services and unused plists'
     'list:List all services managed by brew services'
     'restart:Gracefully restart selected service'
-    'run:Run selected service. Don'"'"'t start at login (nor boot).'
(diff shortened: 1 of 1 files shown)


completions/zsh/_brew_services
```diff
@@ -13,7 +13,7 @@ __brew_services_commands() {
     'cleanup:Get rid of stale services and unused plists'
     'list:List all services managed by brew services'
     'restart:Gracefully restart selected service'
-    'run:Run selected service. Don't start at login (nor boot).'
+    'run:Run selected service. Don'"'"'t start at login (nor boot).'
```
MikeMcQuaid: How about `"run:Run selected service. Don't start at login (nor boot)."`?

artdevjs:
@MikeMcQuaid Agree, looks cleaner

MikeMcQuaid:

PR #110, opened 2017-03-18
Implementation for Linux via systemd

blockloop:
I have been working on an implementation for Linux which uses systemd. Put simply, I moved `cmd/brew-services.rb` to `cmd/launchd.rb` and created `cmd/systemd.rb`. Now the contents of `cmd/brew-services.rb` is as follows 

```ruby
if OS.linux?
  require_relative './systemd'
else
  require_relative './launchd'
end

ServicesCli.run!
```

I created Linuxbrew/brew#300 but was told that Homebrew/brew was accepting upstream changes for Linux. The changes I made here should not affect any existing behavior but should enable the ability to use `brew services` on Linux. 

This will require changes to Homebrew/homebrew-core formula to enable the services by implementing `Formula#systemd` much like `Formula#plist`. Using `respond_to?` I've made it so that `brew services` will skip formulae which do not implement `Formula#systemd`. Ideally we will update Homebrew/homebrew-core formula.rb to have `def systemd` return nil by default (like `def plist`) in the base class, but for now this works.


MikeMcQuaid:
There's an awful lot of duplicated code here. I'm game for this supporting Linux and macOS but I'd like to see abstraction and code reuse first.

blockloop:
There are some pieces that would benefit from reuse and I started out that way, but I wanted to get it working without potential issues in the existing code base first. 

Also, I want to avoid a lot of OS.linux? checks in the code. I plan on creating a ServicesCli base and inheriting from Systemd and Launchd or something like that.

MikeMcQuaid:
> Also, I want to avoid a lot of OS.linux? checks in the code. 

I completely agree 👍 

> I plan on creating a ServicesCli base and inheriting from Systemd and Launchd or something like that.

This sounds sensible. Ideally I'd consider perhaps having more than just the single existing class; this could be a good chance to refactor into multiple classes (and also perhaps files). Thanks!

PR #112, opened 2017-04-06
Convert to using launchctl bootstrap instead of deprecated launchctl load -w

### LINKED ISSUE #111: services start does not detect Invalid property list error
First cut on a formula that installs a plist, and it seems that `brew services start` runs `launchctl load -w` (isn't that deprecated?) and declares success based on `launchctl's` exit status, but ignores errors sent to stderr:

```
pb3:~ jhawk$ brew services start ntp
Unsetting PERL5LIB to protect you from yourself
/Users/jhawk/Library/LaunchAgents/homebrew.mxcl.ntp.plist: Invalid property list
==> Successfully started `ntp` (label: homebrew.mxcl.ntp)
pb3:~ jhawk$ ls -ld /Users/jhawk/Library/LaunchAgents/homebrew.mxcl.ntp.plist
-rw-------  1 jhawk  staff  664 Apr  1 23:38 /Users/jhawk/Library/LaunchAgents/homebrew.mxcl.ntp.plist
pb3:~ jhawk$ cat /Users/jhawk/Library/LaunchAgents/homebrew.mxcl.ntp.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
  <key>KeepAlive</key>
  <true/>
  <key>Label</key>
  <string>homebrew.mxcl.ntp</string>
  <key>ProgramArguments</key>
  <array>
    <string>/usr/local/opt/ntp/sbin/ntpd</string>
	<string>-c</string>
	<string>/private/etc/ntp-restrict.conf</string>
	<string>-g</string>
	<string>-p</string>
	<string>/var/run/ntpd.pid</string>
	<string>-f</string>
	<string>/var/db/ntp.drift</string>
	<string>-l</string>
	<string>/var/log/ntp.log 
  </array>
  <key>RunAtLoad</key>
  <true/>
</dict>
</plist>
pb3:~ jhawk$ launchctl list|grep mxcl
pb3:~ jhawk$ launchctl load -w ~/Library/LaunchAgents/homebrew.mxcl.ntp.plist 
/Users/jhawk/Library/LaunchAgents/homebrew.mxcl.ntp.plist: Invalid property list
pb3:~ jhawk$ echo $?
0
```

```
pb3:cmd jhawk$ git log --format=short brew-services.rb
commit a5115e47b05e8d2a632ba7775599e698b240e5a2
Author: Mike McQuaid <mike@mikemcquaid.com>

    Merge pull request #107 from MikeMcQuaid/fix-run-restart
pb3:cmd jhawk$ 
pb3:cmd jhawk$ brew config
Unsetting PERL5LIB to protect you from yourself
HOMEBREW_VERSION: 1.1.11-226-g974b5e2fa2
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 974b5e2fa26a162d29f97e996eb687b12a46a952
Last commit: 9 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 7ee00ee0a9696f4ad88686fd5a54534d0cfc241b
Core tap last commit: 4 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit broadwell
Homebrew Ruby: 2.0.0-p481
Clang: 7.0 build 700
Git: 2.12.0 => /usr/local/bin/git
Perl: /usr/local/bin/perl => /usr/local/Cellar/perl/5.24.0_1/bin/perl
Python: /usr/bin/python
Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby
Java: 1.8.0_74, 1.6.0_65-b14-468
macOS: 10.10.5-x86_64
Xcode: 7.2.1
CLT: 7.2.0.0.1.1447826929
X11: 2.7.11 => /opt/X11

```

johnhawkinson:
As discussed in #111, convert to using `launchctl bootstrap` (preceded by `launchctl enable`) instead of the deprecated `launchctl load -w` form.

This gives us error reporting, which fixes #111. It also means we're not using a deprecated API, which is good.  And it also paves the way for better separation of root vs. user (e.g. so the same command need not do different things in those cases), and perhaps also to fix the problems of running launchctl under tmux/screen (maybe).

While we're here, document the `-v` flag, which was useful here. (Why doesn't the usage message standout in **bold** like in the builtin commands?)

Example of error detection we didn't have before:

```
bash-3.2# brew services -v run ntp
Unsetting PERL5LIB to protect you from yourself
/bin/launchctl enable system/homebrew.mxcl.ntp
/bin/launchctl bootstrap system /usr/local/opt/ntp/homebrew.mxcl.ntp.plist
/usr/local/Cellar/ntp/4.2.8p10/homebrew.mxcl.ntp.plist: Path had bad ownership/permissions
Error: Failure while executing: /bin/launchctl bootstrap system /usr/local/opt/ntp/homebrew.mxcl.ntp.plist
```

versus

```
bash-3.2# brew services -v run ntp
Unsetting PERL5LIB to protect you from yourself
/bin/launchctl load -w /usr/local/opt/ntp/homebrew.mxcl.ntp.plist
/usr/local/Cellar/ntp/4.2.8p10/homebrew.mxcl.ntp.plist: Path had bad ownership/permissions
==> Successfully started `ntp` (label: homebrew.mxcl.ntp)
```

Also, with a deliberately malformed plist, we used to fail undetectably:

```
bash-3.2# brew services start ntp
Unsetting PERL5LIB to protect you from yourself
/Library/LaunchDaemons/homebrew.mxcl.ntp.plist: Invalid property list
==> Successfully started `ntp` (label: homebrew.mxcl.ntp)
bash-3.2# brew services stop ntp
Unsetting PERL5LIB to protect you from yourself
Error: Service `ntp` is not started.
bash-3.2# 
```

and now we do it right:

```
bash-3.2# brew -v services start ntp
Unsetting PERL5LIB to protect you from yourself
==> Generated plist for ntp:
   <?xml version="1.0" encoding="UTF-8"?>
   <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
   <plist version="1.0">
   <dict>
     <key>KeepAlive</key>
     <true/>
     <key>Label</key>
     <string>homebrew.mxcl.ntp</string>
     <key>ProgramArguments</key>
     <array>
       <string>/usr/local/opt/ntp/sbin/ntpd</string>
       <string>-c</string>
       <string>/private/etc/ntp-restrict.conf</string>
       <string>-n</string>
       <string>-g</string>
       <string>-p</string>
       <string>/var/run/ntpd.pid</string>
       <string>-f</string>
       <string>/var/db/ntp.drift</string>
       <string>-l</string>
       <string>/var/log/ntp.log</stringg>
     </array>
     <key>RunAtLoad</key>
     <true/>
   </dict>
   </plist>
   

/bin/launchctl enable system/homebrew.mxcl.ntp
/bin/launchctl bootstrap system /Library/LaunchDaemons/homebrew.mxcl.ntp.plist
/Library/LaunchDaemons/homebrew.mxcl.ntp.plist: Invalid property list
Error: Failure while executing: /bin/launchctl bootstrap system /Library/LaunchDaemons/homebrew.mxcl.ntp.plist
bash-3.2# 
```

---

Note, this was also discussed, somewhat datedly, at https://github.com/Homebrew/legacy-homebrew/issues/33259.
There was a lot of confusion in that thread, and in particular, there seemed to be the assumption that `launchctl enable` should come after `launchctl load`; that's not the case, and so doesn't work, which lead to proposals of using `launchctl kickstart`. That's unnecessary, and also sounds bad and inelegant :)

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -237,7 +237,7 @@ module ServicesCli
       end
     end
 
-    # "load" a plist, the newfangled way
+    # "load" a plist
     def launchctl_load(plist, _function, service)
       if root?
         domain_target = "system"
@@ -245,12 +245,18 @@ module ServicesCli
         domain_target = "gui/#{Process.uid}"
       end
 
-      safe_system launchctl, "enable", "#{domain_target}/#{service.label}"
-      if $?.to_i.nonzero?
-        odie("Failed to enable `#{service.name}`")
-      end
+      if MacOS.version < :yosemite
+        # This syntax was deprecated in Yosemite
+        safe_system launchctl, "load", "-w", plist
+      else
+        # New syntax has improved errr-handling.
+        safe_system launchctl, "enable", "#{domain_target}/#{service.label}"
+        if $?.to_i.nonzero?
+          odie("Failed to enable `#{service.name}`")
+        end
 
-      safe_system launchctl, "bootstrap", domain_target, plist
+        safe_system launchctl, "bootstrap", domain_target, plist
+      end
       if $?.to_i.nonzero?
         odie("Failed to start `#{service.name}`")
       else
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -1,5 +1,6 @@
-#:  * `services`
-#:    Easily start and stop formulae via launchctl
+#:  * `services` [`-v`|`--verbose`] [list | run | start | stop | restart | cleanup] [...]
+#:    Easily start and stop formulae via launchctl.
+#:    With `-v` or `--verbose`, print more detail.
 #:
 #:    Integrates Homebrew formulae with OS X's `launchctl` manager. Services can be
 #:    added to either `/Library/LaunchDaemons` or `~/Library/LaunchAgents`.
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -237,7 +237,7 @@ module ServicesCli
       end
     end
 
-    # "load" a plist, classically
+    # "load" a plist, the newfangled way
     def launchctl_load(plist, _function, service)
       if root?
         domain_target = "system"
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -237,6 +237,16 @@ module ServicesCli
       end
     end
 
+    # "load" a plist, classically
+    def launchctl_load(plist, function, service)
+        safe_system launchctl, "load", "-w", plist
+        if $?.to_i.nonzero?
+          odie("Failed to start `#{service.name}`")
+        else
+          ohai("Successfully started `#{service.name}` (label: #{service.label})")
+        end
+    end
+
     # Run a service.
     def run(target)
       if target.is_a?(Service) && target.loaded?
@@ -245,13 +255,7 @@ module ServicesCli
       end
 
       Array(target).each do |service|
-        safe_system launchctl, "load", "-w", service.plist
-
-        if $?.to_i.nonzero?
-          odie("Failed to start `#{service.name}`")
-        else
-          ohai("Successfully started `#{service.name}` (label: #{service.label})")
-        end
+        launchctl_load(service.plist, "ran", service)
       end
     end
 
@@ -296,13 +300,7 @@ module ServicesCli
         # Clear tempfile.
         temp.close
 
-        safe_system launchctl, "load", "-w", service.dest.to_s
-
-        if $?.to_i.nonzero?
-          odie("Failed to start `#{service.name}`")
-        else
-          ohai("Successfully started `#{service.name}` (label: #{service.label})")
-        end
+        launchctl_load(service.dest.to_s, "started", service)
       end
     end
 
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -253,7 +253,6 @@ module ServicesCli
       end
 
       ohai("Successfully #{function} `#{service.name}` (label: #{service.label})")
-
     end
 
     # Run a service.
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -246,20 +246,14 @@ module ServicesCli
 
       if MacOS.version >= :yosemite
         safe_system launchctl, "enable", "#{domain_target}/#{service.label}"
-        if $?.to_i.nonzero?
-          odie("Failed to enable `#{service.name}`")
-        end
-
         safe_system launchctl, "bootstrap", domain_target, plist
       else
         # This syntax was deprecated in Yosemite
         safe_system launchctl, "load", "-w", plist
       end
-      if $?.to_i.nonzero?
-        odie("Failed to start `#{service.name}`")
-      else
-        ohai("Successfully started `#{service.name}` (label: #{service.label})")
-      end
+
+      ohai("Successfully started `#{service.name}` (label: #{service.label})")
+
     end
 
     # Run a service.
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -244,16 +244,16 @@ module ServicesCli
         domain_target = "gui/#{Process.uid}"
       end
 
-      if MacOS.version < :yosemite
-        # This syntax was deprecated in Yosemite
-        safe_system launchctl, "load", "-w", plist
-      else
+      if MacOS.version >= :yosemite
         safe_system launchctl, "enable", "#{domain_target}/#{service.label}"
         if $?.to_i.nonzero?
           odie("Failed to enable `#{service.name}`")
         end
 
         safe_system launchctl, "bootstrap", domain_target, plist
+      else
+        # This syntax was deprecated in Yosemite
+        safe_system launchctl, "load", "-w", plist
       end
       if $?.to_i.nonzero?
         odie("Failed to start `#{service.name}`")
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -237,7 +237,6 @@ module ServicesCli
       end
     end
 
-    # "load" a plist
     def launchctl_load(plist, _function, service)
       if root?
         domain_target = "system"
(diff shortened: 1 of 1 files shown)


ilovezfs:
Note that this will break backwards compatibility on earlier versions of macOS.

johnhawkinson:
Did I misunderstand @MikeMcQuaid:

> That seems like a good idea. My concern is that it may limit the backwards compatibility of brew services but 10.10 and above seems fine at this point.

It's easy enough to conditionalize this on `MacOS.version >= :yosemite` and keep the old code, it's like 1 line plus the if/else (total 3). What's desired?

ilovezfs:
Of course you didn't misunderstand. To be explicit: i don't think we should break backward compatibility if it can be avoided.

MikeMcQuaid:
> It's easy enough to conditionalize this on MacOS.version >= :yosemite and keep the old code, it's like 1 line plus the if/else (total 3). What's desired?

That would be good. I thought the changes would be more involved so it should be relatively straightforward to keep the old code around, thanks (and sorry to change my mind).

johnhawkinson:
OK, slightly more as I added 2 comment lines.
I assume it's OK to break the `safe_system` / `if $?.to_i_.nonzero?` pair across the `if` block like that, but if not of course we can add more code duplication…

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
       end
     end
 
+    # "load" a plist
+    def launchctl_load(plist, _function, service)
```
MikeMcQuaid: Skip `_function` if it's never used.

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
       end
     end
 
+    # "load" a plist
```
MikeMcQuaid: Can skip this comment, pretty self-explanatory.

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
...
+        domain_target = "gui/#{Process.uid}"
+      end
+
+      if MacOS.version < :yosemite
+        # This syntax was deprecated in Yosemite
+        safe_system launchctl, "load", "-w", plist
+      else
+        # New syntax has improved errr-handling.
+        safe_system launchctl, "enable", "#{domain_target}/#{service.label}"
+        if $?.to_i.nonzero?
```
MikeMcQuaid: `safe_system` will fail with an error so this failure check will never be hit (I realise the previous code also made this mistake).

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
...
+          odie("Failed to enable `#{service.name}`")
+        end
+
+        safe_system launchctl, "bootstrap", domain_target, plist
+      end
+      if $?.to_i.nonzero?
+        odie("Failed to start `#{service.name}`")
```
MikeMcQuaid: `safe_system` will fail with an error so this failure check will never be hit (I realise the previous code also made this mistake).

ilovezfs:
>OK, slightly more as I added 2 comment lines.

We're sorry. You have exceeded your line quota. Please try again next time.

cmd/brew-services.rb
MikeMcQuaid: But it looks like it probably wants to be called `function` and used in `ohai("Successfully started `#{service.name}` (label: #{service.label})")`

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
...
 
+    # "load" a plist
+    def launchctl_load(plist, _function, service)
+      if root?
+        domain_target = "system"
+      else
+        domain_target = "gui/#{Process.uid}"
+      end
+
+      if MacOS.version < :yosemite
```
MikeMcQuaid: Pedantry but make this the `else` so the first block people see is the current version.

cmd/brew-services.rb
```diff
@@ -236,6 +237,33 @@ def restart(target)
...
+        # This syntax was deprecated in Yosemite
+        safe_system launchctl, "load", "-w", plist
+      else
+        # New syntax has improved errr-handling.
```
MikeMcQuaid: Can remove this comment.

cmd/brew-services.rb
johnhawkinson: That was the intention, but I was concerned that changing the message from "started" run "ran" in the `brew run` case might break callers. So I wanted to make the API for the function call support that concept without commiting to it. Which way should I go? Is it OK to change the output for the `brew run` case?

(On the other hand, maybe `brew run` really does `start` things instead of `running` them. In which case this should all just go away?)

MikeMcQuaid: > Is it OK to change the output for the brew run case?

Yes.



> (On the other hand, maybe brew run really does start things instead of running them. In which case this should all just go away?)

That was to preserve backwards compatibility of commands.

johnhawkinson:
OK, I think that's all of them?

MikeMcQuaid:
@johnhawkinson Test seems to be failing; looks like `mysql` didn't come up.

johnhawkinson:
> @johnhawkinson Test seems to be failing; looks like mysql didn't come up.

Wha-huh? That seems so unlike---oh, how convenient. Final commit was cd4fc93 which changed only whitespace (removed a blank line). Prior commit also had a Travis run that succeeded. I think this means there's some nondeterminance in the mysql travis test (consistent with the "sleep 2" crap)?

I suppose the other possibility is there is nondeterminancy in the code in this PR, but that does not seem likely to me, since I didn't ever have it fail through repeated testing and whacking on my local system.

Thoughts? "It's never the test…except when it is."

MikeMcQuaid:
Will rerun the test to see if it's random. If it is we might just bump the `sleep` a bit (proper engineering, that :trollface:)

MikeMcQuaid:
Yeh, passed second time. Bumping sleep.

ilovezfs:
:sleeping:

MikeMcQuaid:
Local tests look good to me. Nice work again here @johnhawkinson!

ilovezfs:
:tada:

PR #117, opened 2017-05-21
brew-services: use non-deprecated unload methods.

MikeMcQuaid:
Use `launchctl`'s `kill` and `disable` subcommands rather than `unload` to avoid weirdness where services aren't properly terminated.

CC @bryanaknight who ran into this.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -70,6 +70,7 @@ module ServicesCli
 
     # Find all currently running services via launchctl list.
     def running
+      # TODO: find replacement for deprecated "list"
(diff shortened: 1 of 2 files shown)


cmd/brew-services.rb
```diff
@@ -326,7 +328,13 @@ def stop(target)
       Array(target).select(&:loaded?).each do |service|
         if service.dest.exist?
           puts "Stopping `#{service.name}`... (might take a while)"
-          safe_system launchctl, "unload", "-w", service.dest.to_s
+          if MacOS.version >= :yosemite
+            safe_system launchctl, "kill", "TERM", "#{domain_target}/#{service.label}"
```
JCount: I believe this needs to be either `"SIGTERM"` or `"-15"`.

PR #126, opened 2017-06-15
brew-services: follow Apple guidance for launchd plist file perms

mcashwell:
Files such as [~]/Library/LaunchDaemons/foobar.plist installed by
[sudo] brew services start foobar
ended up with overly restrictive 0600 permissions. While launchd
accepted this because it's not a security risk, other system daemons
run as non-root and need world-read on those files to work properly.


MikeMcQuaid:
Thanks for your first contribution to Homebrew Services, @mcashwell! Without people like you submitting PRs we couldn't run this project. You rock!

PR #149, opened 2018-02-07
Rename bash completion script

jasonkarns:
On my system, the brew-services completions were not running. The script was being registered to bash-completion, but was then presumably being overwritten by homebrew core's registration.

This completion script must wrap homebrew core's completion function (_brew) in order to work (since it's adding completions for subcommands of brew). As such, homebrew core's completion script must be loaded first. The bash-completion package (at least, v2) seems to load them in globbed file order. The homebrew core completion script is named `brew`. Thus, in order for this completion script to be loaded after core, it must sort _after_ `brew` in a listing. Dropping the leading underscore pushes this script after brew core. Also, swapped the mid-word underscore for a hyphen to match the style that brew-cask uses/used.


MikeMcQuaid:
Seems reasonable, thanks @jasonkarns!

PR #160, opened 2018-05-03
Sort "brew services list" based on service name

stefansundin:
I am not sure, but I seem to remember this being sorted before. I _think_ that on HFS+, directories always returned directory listing results in sorted order, but that changed on APFS. I had the same issue in another project of mine.

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -118,7 +118,7 @@ module ServicesCli
 
   # List all available services with status, user, and path to the plist file.
   def list
-    formulae = available_services.map do |service|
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @stefansundin!

stefansundin:
Does it still count as my first contribution even though I made https://github.com/Homebrew/brew/pull/4022? :D

MikeMcQuaid:
It was to this repo so you get congratulated again 😉

PR #183, opened 2019-03-11
error when using brew services on linux

morgo:
This patch is related to https://github.com/Homebrew/brew/pull/5864 and fixes  https://github.com/Homebrew/brew/issues/5847


MikeMcQuaid:
Thanks for this @morgo, great work!

PR #191, opened 2019-06-22
brew services: Add pattern flag to be able to select mutliple formulars in one command

nicolas-heinig:
Hello there!

I thought it would be nice to be able to select formulars with a pattern.
That enables us to the following things:

1. Select multiple formulars in one command: 
  `brew services stop -p '(mysql|memcached)'` to stop mysql and memcached in one command.
2. Select a version specific fomular without the need to type the version to it:
  `brew services stop -p mysql` when you only have `mysql@5.7` installed.

Let me know what you think of my idea and implementation.

Thank you very much,
Nick

### DIFF
--- a/cmd/brew-services.rb
+++ b/cmd/brew-services.rb
@@ -4,25 +4,26 @@
 #:
 #:  Manage background services with macOS' `launchctl`(1) daemon manager
 #:
+#:       -p, --pattern                   run <subcommand> on all services matching the pattern.
 #:       --all                           run <subcommand> on all services.
 #:
 #:  [`sudo`] `brew services list`
 #:
 #:  List all running services for the current user (or root).
 #:
-#:  [`sudo`] `brew services run` (<formula>|`--all`)
+#:  [`sudo`] `brew services run` (<formula>|`-p` <pattern>|`--all`)
 #:
 #:  Run the service <formula> without registering to launch at login (or boot).
 #:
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> 1. `brew services stop -p '(mysql|memcached)'` to stop mysql and memcached in one command.

I've not seem anyone previously request that. What's the advantage of doing it in a single command?

> 2\. Select a version specific fomular without the need to type the version to it:
>     `brew services stop -p mysql` when you only have `mysql@5.7` installed.

I definitely think this is undesirable; you could end up stopping the wrong service due to a typo.

---

Thanks for the PR anyway but I'm unsure we're going to merge this.

nicolas-heinig:
> I've not seem anyone previously request that. What's the advantage of doing it in a single command?
I found myself often in the position where I would need to stop two services. After stoping the first one I would need to type the command again. You could have the same convenience when you are able to provide more than one formular name to the services commands. What do you think of this idea?

nicolas-heinig:
>  You could end up stopping the wrong service due to a typo.

Hmm. Yes you are right, that's a danger of the pattern. Did you thought about adding aliasing for formulars? So that when I only run `mysql@5.7` I could reference it with simply `mysql` ?

MikeMcQuaid:
Sorry @nicolas-heinig, passing on this.

PR #220, opened 2020-05-02
services: graceful handle program without slashes

### LINKED ISSUE #219: Starting services as root with first argument "sh" does not work
**What did I try:**
I wanted to start `node_exporter` as root to start it on boot time.
`sudo brew services start node_exporter`
**What did I expect:**
Node exporter started and added to services started at boot and running.
**What happend:**
`Error: No such file or directory @ realpath_rec - <cwd>/sh`
**What I did to debug this issue:**
I created the file that was missing by `touch` and then saw that `brew services` was changing the ownership of `<cwd>/sh` and `<cwd>` to `root:admin`. After looking a little bit into the files I found that in the formula (https://github.com/Homebrew/homebrew-core/blob/master/Formula/node_exporter.rb) in the plist the first argument is sh to use a file to get the arguments for `node_exporter`. After looking a little bit into this repository I found this lines (https://github.com/Homebrew/homebrew-services/blob/master/lib/services_cli.rb#L305) where the first argument is extracted to change ownership.

So I do not know if this is a issue in the formula but the error message could have been more meaningful. If plist files with first arguments like this are valid then they should be excluded from ownership change I think. Also a warning can inform the user that such plist files are not supported and he has to manually change ownership for relevant files followed by the list of arguments. Unfortunately I do not know Ruby so I could only create a bad quality PR I guess.

schribl:
Closes #219 

**Problem:**
Described in Issue #219 
**Change:**
After looking into `man launchd.plist` and `man execv` the first entry in `ProgramArguments` does not have to contain an absolute path if `Program` property is missing. So in the change the first argument to `execv`/`execvp` is not considered as a file to change ownership if it does not contain a slash. To warn the user about possible issues some disclaimer is printed as warning.

### DIFF
--- a/lib/services_cli.rb
+++ b/lib/services_cli.rb
@@ -301,13 +301,36 @@ module Homebrew
           end
           next unless plist
 
+          # From man launchd.plist and man execv:
+          # The first argument for execv/exevp is either the Property "Program"
+          # or the first entry of "ProgramArguments".
+          program_location = if plist["Program"].present?
+            plist["Program"]
+          elsif plist["ProgramArguments"]&.first.present?
+            plist["ProgramArguments"]&.first
+          end
+
           root_paths = []
-          if (program_location = plist["ProgramArguments"]&.first)
-            program_location_path = Pathname(program_location).realpath
-            root_paths += [
-              program_location_path,
-              program_location_path.parent.realpath,
-            ]
+          if program_location
+            # If "Program" is missing and first entry of "ProgramArguments" does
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @schribl! Will pass on this PR but thanks for creating it; I understand the problem now and have made three PRs to address it in different ways: https://github.com/Homebrew/homebrew-services/pull/221

schribl:
I commented some notes I am able to give on your PRs @MikeMcQuaid

PR #223, opened 2020-05-21
Enable custom_plist with restart

rcrews:
Love the custom plist support. This request is to extend the feature to the restart command. Took a long time to discover how to use custom plists. Happy to describe it in the documentation too, if you can point me to what doc files should be edited.

### DIFF
--- a/lib/services_cli.rb
+++ b/lib/services_cli.rb
@@ -79,7 +79,7 @@ module Homebrew
       case subcommand.presence
       when nil, "list", "ls" then list
       when "cleanup", "clean", "cl", "rm" then cleanup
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @rcrews!



> Happy to describe it in the documentation too, if you can point me to what doc files should be edited.

Would love this. The main places are the README and the usage banner: https://github.com/Homebrew/homebrew-services/blob/a64759cad5c952f49f52eece50f0790b00e38182/cmd/services.rb#L10-L35

PR #278, opened 2020-11-16
Remove Unmatched Quote in HomeBrew Services CLI Message

ryanjbonnell:
When using HomeBrew Services to stop a command originally run as `root`, a suggestion appears if `sudo` isn't specified.

Consider the following screenshot as an example:

Screen Shot 2020-11-16 at 2 50 28 PM

As illustrated above, an accidental double quote is shown at the end of the status message `sudo brew services stop php@7.2"`.

This change proposes to fix the output of the message to omit the trailing quote.

### DIFF
--- a/lib/services_cli.rb
+++ b/lib/services_cli.rb
@@ -365,7 +365,7 @@ module Homebrew
         if target.started?
           odie <<~EOS
             Service `#{target.name}` is started as `#{target.started_as}`. Try:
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @ryanjbonnell!

PR #280, opened 2020-11-23
bash-completion: separate out running commands

bradschwartz:
The existing shell completion only found plist files that had been started, which means you had to know the name of the service to start it. This changes it so only the `stop|cleanup` uses this.
Commands that run services (`start|restart|run`) now find services using:
`find "${BREW_PREFIX}/Cellar"  -mindepth 3 -maxdepth 3 -name 'homebrew.mxcl.*.plist'`.
Where BREW_PREFIX is set on shell startup as `brew --prefix` output.

Adds active services started with `sudo` to the completions as well.

On my machine calling `brew --prefix` adds ~30ms latency to shell startup, but no impact after that. If there's a better way to find that would love to know!

I have a version that gets rid of the duplicate string expressions but it was pretty difficult to grok, going with this instead for readability.

### DIFF
--- a/completions/bash/brew-services
+++ b/completions/bash/brew-services
@@ -2,14 +2,24 @@
 # brew services completion
 #########################################################################
 
+# Called once on shell startup
+BREW_PREFIX="$(brew --prefix)"
+
 # Complete brew services xxxxxxx
 _brew_services() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
   local prv="$(__brewcomp_prev)"
   case "$prv" in
-    start|stop|restart|cleanup)
+    stop|cleanup)
       # get list of agent plists, clean up with param expansion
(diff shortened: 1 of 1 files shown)


completions/bash/brew-services
```diff
@@ -2,14 +2,24 @@
 # brew services completion
 #########################################################################
 
+# Called once on shell startup
+BREW_PREFIX="$(brew --prefix)"
```
MikeMcQuaid, line 6:
```suggestion
HOMEBREW_PREFIX="$(brew --prefix)"
```

MikeMcQuaid (commented):
This is looking good, nice work! One minor tweak.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @bradschwartz!

PR #355, opened 2021-06-09
Add restart task to test workflow

keeran:
Adding a `restart` task to catch anything similar to #354 in the future :)

/cc @issyl0 @SMillerDev


issyl0 (approved):
How strange that we didn't test this before. Thanks so much for your first contribution to Homebrew, @keeran!

MikeMcQuaid:
You rock @keeran!

PR #391, opened 2021-11-01
fix: --all flag causing a unknown method `available_services` error

jimeh:
Commit 7c4a420 moved the available_services class method from
Service::ServicesCli to Service::Formulae, but cmd/service.rb still
referred to the old class, causing a unknown method error.

### DIFF
--- a/cmd/services.rb
+++ b/cmd/services.rb
@@ -71,7 +71,7 @@ module Homebrew
     end
 
     target = if args.all?
-      Service::ServicesCli.available_services
+      Service::Formulae.available_services
     elsif formula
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):
I thought I got them all, thanks @jimeh !

jimeh:
@SMillerDev no worries, happens to the best of us :)... I'd imagine the `--all` flag sees little use in general anyway. I only noticed cause I'm messing with a xbar plugin to show/manage brew services, and it has start/stop/restart all commands which I wanted to make sure actually worked.

PR #439, opened 2022-01-16
Fix `brew services list` columns

Rylan12:
I noticed that in certain situation, `brew services list` doesn't property column-ify its output:

```console
$ brew services list
Name      Status  User         File
mosquitto started rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.mosquitto.plist
stubby    error  256 rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.stubby.plist
unbound   none
znc       scheduled rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.znc.plist
```

This is because it assumes that the longest possible status string is 7 characters. This is no longer true, though, since `scheduled` and `error 256` are both longer.

I rewrote the method that outputs these columns to check for the longest status column and pad every other item to that width. This is complicated a little bit by the color codes from `Tty.color` and `Tty.reset`. These characters are counted toward the cutoff number in the format string, but aren't rendered. I think the way I did it makes sense, but I'm definitely open to suggestions if it's not clear enough.

The new output:

```console
$ brew services list
Name      Status     User         File
mosquitto started    rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.mosquitto.plist
stubby    error  256 rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.stubby.plist
unbound   none
znc       scheduled  rylanpolster ~/Library/LaunchAgents/homebrew.mxcl.znc.plist
```

### DIFF
--- a/lib/service/commands/list.rb
+++ b/lib/service/commands/list.rb
@@ -21,21 +21,30 @@ module Service
       # Print the table in the CLI
       # @private
       def print_table(formulae)
-        longest_name = [formulae.max_by { |formula| formula[:name].length }[:name].length, 4].max
-        longest_user = [formulae.map { |formula| formula[:user].nil? ? 4 : formula[:user].length }.max, 4].max
-
-        headers = "#{Tty.bold}%-#{longest_name}.#{longest_name}<name>s %-7.7<status>s " \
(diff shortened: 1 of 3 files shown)


SMillerDev (approved):

MikeMcQuaid:
Nice catch @Rylan12!

PR #441, opened 2022-01-17
services: fix missing highlight

gromgit:
This makes clear that `start` and `run` are command literals, not action descriptions.

Addresses https://github.com/Homebrew/homebrew-core/issues/93285.


MikeMcQuaid (approved):

Rylan12 (approved):
Makes sense. Thanks, @gromgit!

PR #443, opened 2022-01-20
lib/service/commands/list: only output "no services" warning with TTY.

### LINKED ISSUE #438: Warning: No services available to control with `brew services`
## Description

My issue is related to https://github.com/Homebrew/homebrew-services/issues/218, but this is happening on a clean Mac running macOS Monterey v12.1. I had just reinstalled some of my packages using [this Brewfile](https://github.com/jvanderen1/brewfile/commit/9edc324d900a4a6b68d70c6bd1920ca08617fe31), but then added some after successfully installing them.

## Reproduction

I executed the following command:

```shell
brew bundle dump
```

Then, I received the following message:

```shell
Warning: No services available to control with `brew services`
```

However, the `Brewfile` was successfully generated.

## Investigation

This message appears to be coming from [these lines](https://github.com/Homebrew/homebrew-services/blob/master/lib/service/commands/list.rb#L12-L19) of code, but it's unclear to me why this line is invoked after running `brew bundle dump`. I'd need to investigate more.

MikeMcQuaid:
This avoids it being output when using `brew bundle` or other tools that parse `brew services` output.

Fixes #438

### DIFF
--- a/lib/service/commands/list.rb
+++ b/lib/service/commands/list.rb
@@ -11,8 +11,8 @@ module Service
 
       def run
         formulae = Formulae.services_list
-        if formulae.empty?
-          opoo("No services available to control with `#{Service::ServicesCli.bin}`")
+        if formulae.blank?
+          opoo "No services available to control with `#{Service::ServicesCli.bin}`" if $stderr.tty?
           return
         end
         print_table(formulae)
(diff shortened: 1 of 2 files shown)


jvanderen1:
Thanks @MikeMcQuaid! Lgtm

PR #468, opened 2022-03-24
Removed #launchctl-kill and refactored #stop in services_cli.rb

apainintheneck:
This PR is related to something that came up during [the review](https://github.com/Homebrew/homebrew-services/pull/464#discussion_r824053818) of the last one related to adding the kill command. Essentially, the previous kill function was renamed to launchctl_kill in ServicesCli to make room for the new method but it duplicated a lot of the functionality of the old one. This PR removes that method and then refactors the methods that called it.

### ServicesCli#launchctl_kill was used by two methods
1. #kill_orphaned_services
	It was replaced by the new #kill method
2. #stop
	It was replaced by the launchctl stop command

### The #stop method
I also ended up refactoring the ServicesCli#stop method to remove the call to launchctl_kill and make it more consistent with documentation.
- remove unnecessary checks
- added a call to the systemctl disable command so that service
  won't start up again after a restart
- moved the removal of the service file to the top of the method
  because it's universal
- added a message if the service can't be stopped

Let me know if you have any questions.

### DIFF
--- a/lib/service/services_cli.rb
+++ b/lib/service/services_cli.rb
@@ -39,15 +39,12 @@ module Service
       cleaned = []
       running.each do |label|
         if (svc = FormulaWrapper.from(label))
-          unless svc.dest.file?
-            puts format("%-15.15<name>s #{Tty.bold}stale#{Tty.reset} => killing service...", name: svc.name)
-            launchctl_kill(svc)
-            cleaned << label
-          end
+          cleaned << label unless svc.dest.file?
         else
           opoo "Service #{label} not managed by `#{bin}` => skipping"
         end
       end
+      kill(cleaned)
       cleaned
     end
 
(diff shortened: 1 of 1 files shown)


apainintheneck:
Well, clearly the refactor of the stop command didn't go well. I'll take a look at it again later.

MikeMcQuaid (commented):
Looks good so far. Shout when CI is passing!

apainintheneck:
After playing around with it, I was able to get it working. The hangup was on Linux. I did the following to get it working.
1. Changed out `systemctl stop` command for `systemctl disable --now` which disables and stops the service.
2. Stopped checking the loaded? command before deciding that the service has been stopped. (This might not be relevant and I could try adding it back in.)
3. Stopped deleting the systemctl service file which seemed to cause problems when restarting. Has this come up before? If so, it would probably be a good idea to add a comment explaining that the service file shouldn't get removed when using systemctl.

SMillerDev (approved):
Looks good to me @apainintheneck, thanks  ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a bit better for homebrew users around the world! 👍 🎉

PR #478, opened 2022-04-29
Restore old restart command

apainintheneck:
This restores the old restart command as discussed so that it can be used to update service files after upgrading formulas. I added a note to the `restart.rb` file mentioning the intended behavior but I'm not exactly sure how to add tests so this sort of thing doesn't happen again in the future.

The first thing that comes to mind would be to add a functional test that checks to make sure that the service file was updated/overwritten after the call to `brew services restart`. Finding a cross platform way to do that and reconciling the different locations of the services files on macOS and linux is probably beyond me to be honest.

Either way this PR should fix the immediate problem.

### DIFF
--- a/lib/service/commands/restart.rb
+++ b/lib/service/commands/restart.rb
@@ -5,6 +5,11 @@ module Service
     module Restart
       module_function
 
+      # NOTE: The restart command is used to update service files
+      # after a package gets updated through `brew upgrade`.
+      # This works by removing the old file with `brew services stop`
+      # and installing the new one with `brew services start|run`.
+
       TRIGGERS = %w[restart relaunch reload r].freeze
 
       def run(targets, custom_plist, verbose:)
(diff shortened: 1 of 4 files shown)


SMillerDev (approved):

lib/service/commands/restart.rb
```diff
@@ -5,29 +5,32 @@ module Commands
...
-            opoo "Unable to restart `#{service.name}` (label: #{service.service_name})"
+            # group not-started services with started ones for restart
+            started << service
           end
+          ServicesCli.stop([service]) if service.loaded?
         end
 
-        ServicesCli.start(start_targets, verbose: verbose) if start_targets.present?
+        ServicesCli.run(ran) unless ran.empty?
+        ServicesCli.start(started) unless started.empty?
```
MikeMcQuaid, line 33:
```suggestion
        ServicesCli.run(ran) if ran.present?
        ServicesCli.start(started) if started.present?
```

MikeMcQuaid (approved):
Great work! One optional suggestion.

PR #500, opened 2022-08-11
systemd: support systemctl --user without dbus

### LINKED ISSUE #496: Linux (debian): requires dbus-session-bus or should retain XDG_RUNTIME_DIR for systemctl --user to function
I'm using linuxbrew on Debian 10, which has systemd with `--user` support. However, none of the `brew services` commands actually work.

I tracked this down to the `XDG_RUNTIME_DIR` variable not being set; `/home/linuxbrew/.linuxbrew/Homebrew/bin/brew` has a whitelist of environment variables it keeps (unless the deprecated `HOMEBREW_NO_ENV_FILTERING` env var is set), and this list doesn't include `XDG_RUNTIME_DIR`.

The variable is crucial however, omitting it causes `systemctl --user` commands to fail:

```sh
$ env -u XDG_RUNTIME_DIR /bin/systemctl --user status
Failed to connect to bus: No such file or directory
$ /bin/systemctl --user status
● <hostname>
    State: running
     Jobs: 0 queued
   Failed: 0 units
    Since: Tue 2022-08-02 19:30:48 BST; 21h ago
   CGroup: /user.slice/user-<uid>.slice/user@<uid>.service
           └─init.scope
             ├─<pid> /lib/systemd/systemd --user
             └─<pid> (sd-pam)
```

(I replaced some elements with `<...>` placeholders for security reasons).

I can get `brew services` commands to work by editing the `.../bin/brew` command and adding `XDG_RUNTIME_DIR` to the whitelist. Perhaps this is something that should be made permanent?

I reported this _here_ because this issue affects the `brew services` commands only, but perhaps should be reported in the certral core repo instead.

mjpieters:
Pass along the `XDG_RUNTIME_DIR` so systemctl --user can fall back to it's
internal dbus socket; needed when no user-session dbus daemon is installed.

Depends on Homebrew/brew#13686 to be merged for `HOMEBREW_XDG_RUNTIME_DIR` to
exist.

Fixes #496

### DIFF
--- a/cmd/services.rb
+++ b/cmd/services.rb
@@ -93,6 +93,7 @@ module Homebrew
 
     if Service::System.systemctl?
       ENV["DBUS_SESSION_BUS_ADDRESS"] = ENV.fetch("HOMEBREW_DBUS_SESSION_BUS_ADDRESS", nil)
+      ENV["XDG_RUNTIME_DIR"] = ENV.fetch("HOMEBREW_XDG_RUNTIME_DIR", nil)
     end
 
     # Dispatch commands and aliases.


dawidd6 (approved):

MikeMcQuaid:
Merging despite this not being tagged yet because it'll be a no-op until the tag anyway.

Great work @mjpieters!

PR #536, opened 2023-01-14
formula_wrapper: return hash if there is no service command

dawidd6:
Companion PR to https://github.com/Homebrew/brew/pull/14371.

### DIFF
--- a/lib/service/formula_wrapper.rb
+++ b/lib/service/formula_wrapper.rb
@@ -190,6 +190,9 @@ module Service
       return hash unless service?
 
       service = load_service
+
+      return hash if service.command.blank?
+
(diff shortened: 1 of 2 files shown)


SMillerDev (approved):

lib/service/formula_wrapper.rb
```diff
@@ -190,6 +190,9 @@ def to_hash
       return hash unless service?
```
MikeMcQuaid, line 190:
```suggestion
      return hash if service.blank?
```

MikeMcQuaid (approved):
Looks good! Tiny suggestion based on consistency for existing code.

PR #551, opened 2023-02-15
Bash completions: add __brewcomp_prev

maxim-belkin:
Previously, `__brewcomp_prev` function was provided in Homebrew/brew Bash completion script. It was removed in Homebrew/brew#14545 because no other function in Homebrew/brew was using it. Since Bash completions for Homebrew/services relies on this function, we restore `__brewcomp_prev` here.

Addresses Homebrew/homebrew-services#549

### DIFF
--- a/completions/bash/brew-services
+++ b/completions/bash/brew-services
@@ -5,10 +5,20 @@
 # Called once on shell startup
 : "${HOMEBREW_PREFIX:=$(brew --prefix)}"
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks again @maxim-belkin!

maxim-belkin:
Thanks for merging this in, Mike!

PR #588, opened 2023-09-27
Support (and set) `launchd` user domain target

MikeMcQuaid:
When running `brew services` over `ssh` or `sudo` as a non-`root` user the `gui` domain target for `launchd` does not work as expected. This produces confusing failures for users.

It was tried before to detect if this was needed from the `.plist` or changing globally but both approaches were unsuccessful/error-prone.

Instead, use the `user` domain automatically on the cases we know it's needed: when running over `ssh` or through `sudo`. To make clear to users what's happening in these cases: output a warning (which can be hidden with an output environment variable).

For this to work, `launchctl list` is no longer sufficient. The output here, even when run as `root`, does not properly list `user` domain services. Instead, we need to use `launchctl print` to correctly query the status of these services. This also seems to correctly handle some `launchd` edge-cases where launched services cannot be detected and lets `brew services` now stop them.

While we're here, fix some related issues I came upon while working on this:
- improve the `brew services` command documentation to note it now runs on Linux/`systemd` too
- use `named_args` and remove the `custom_plist` deprecation: it's been long enough and this cleans up the code nicely.
- if a service is status `:other` (an edge-case that shouldn't be possible in normal operation): actually output this rather than failing with a `nil` error
- avoid running `launchctl list` multiple times when unnecessary
- `brew services --debug` now outputs the raw output from `launchctl` or `systemctl` to aid debugging/development
- cleanup some code style and avoid use of `plist?`
- add `System.systemctl_args` to avoid repeating `System.systemctl` and `System.systemctl_scope` in every call site

--- 

Depends on https://github.com/Homebrew/brew/pull/16041 to detect `sudo` or `ssh` correctly.

### DIFF
--- a/cmd/services.rb
+++ b/cmd/services.rb
@@ -10,13 +10,15 @@ module Homebrew
       usage_banner <<~EOS
         `services` [<subcommand>]
 
-        Manage background services with macOS' `launchctl`(1) daemon manager.
+        Manage background services with macOS' `launchctl`(1) daemon manager or
+        Linux's `systemctl`(1) service manager.
 
-        If `sudo` is passed, operate on `/Library/LaunchDaemons` (started at boot).
-        Otherwise, operate on `~/Library/LaunchAgents` (started at login).
+        If `sudo` is passed, operate on `/Library/LaunchDaemons`/`/usr/lib/systemd/system`  (started at boot).
+        Otherwise, operate on `~/Library/LaunchAgents`/`~/.config/systemd/user` (started at login).
 
-        [`sudo`] `brew services` [`list`] (`--json`):
+        [`sudo`] `brew services` [`list`] (`--json`) (`--debug`):
         List information about all managed services for the current user (or root).
+        Provides more output from Homebrew and `launchctl`(1) or `systemctl`(1) if run with `--debug`.
 
         [`sudo`] `brew services info` (<formula>|`--all`|`--json`):
         List all managed services for the current user (or root).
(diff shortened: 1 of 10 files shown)


SMillerDev (approved):

Uning456789:
> When running `brew services` over `ssh` or `sudo` as a non-`root` user the `gui` domain target for `launchd` does not work as expected. This produces confusing failures for users.
> 
> It was tried before to detect if this was needed from the `.plist` or changing globally but both approaches were unsuccessful/error-prone.
> 
> Instead, use the `user` domain automatically on the cases we know it's needed: when running over `ssh` or through `sudo`. To make clear to users what's happening in these cases: output a warning (which can be hidden with an output environment variable).
> 
> For this to work, `launchctl list` is no longer sufficient. The output here, even when run as `root`, does not properly list `user` domain services. Instead, we need to use `launchctl print` to correctly query the status of these services. This also seems to correctly handle some `launchd` edge-cases where launched services cannot be detected and lets `brew services` now stop them.
> 
> While we're here, fix some related issues I came upon while working on this:
> 
> * improve the `brew services` command documentation to note it now runs on Linux/`systemd` too
> * use `named_args` and remove the `custom_plist` deprecation: it's been long enough and this cleans up the code nicely.
> * if a service is status `:other` (an edge-case that shouldn't be possible in normal operation): actually output this rather than failing with a `nil` error
> * avoid running `launchctl list` multiple times when unnecessary
> * `brew services --debug` now outputs the raw output from `launchctl` or `systemctl` to aid debugging/development
> * cleanup some code style and avoid use of `plist?`
> * add `System.systemctl_args` to avoid repeating `System.systemctl` and `System.systemctl_scope` in every call site
> 
> Depends on [Homebrew/brew#16041](https://github.com/Homebrew/brew/pull/16041) to detect `sudo` or `ssh` correctly.

PR #593, opened 2023-10-05
Add `--no-wait` argument

MikeMcQuaid:
This prevents us from looping and waiting for `brew services stop` to finish stopping the service if we just want to attempt once and immediately exit.

While we're here, also fix verbose output for `brew services restart`.

### DIFF
--- a/lib/service/services_cli.rb
+++ b/lib/service/services_cli.rb
@@ -124,7 +124,7 @@ module Service
     end
 
     # Stop a service and unload it.
-    def stop(targets, verbose: false)
+    def stop(targets, verbose: false, no_wait: false)
       targets.each do |service|
         unless service.loaded?
           rm service.dest if service.dest.exist? # get rid of installed service file anyway, dude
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
@SMillerDev Good catch, added.

lib/service/services_cli.rb
```diff
@@ -142,14 +142,23 @@ def stop(targets, verbose: false)
           next
         end
 
-        puts "Stopping `#{service.name}`... (might take a while)"
+        systemctl_args = System.systemctl_args
+        if no_wait
+          systemctl_args << "--no-block"
```
apainintheneck, line 147: It doesn't look like the `systemctl_args` variable is actually used anywhere in this method outside of lines 145 and 147. I assume it's supposed to be used on line 154 as well.

MikeMcQuaid: @apainintheneck Whoops, opened PR in https://github.com/Homebrew/homebrew-services/pull/596

PR #599, opened 2023-10-28
Use `plist` over `rexml`

Bo98:
REXML is disappearing in Ruby 3 and will move to Homebrew/brew's Gemfile. At this time there are no plans to commit it to the repository as a vendored gem and `brew services` should work without needing to install any extra gems. `plist` is however already there as a committed vendored gem so we can use that. It also has better API.

I haven't tested the `--sudo-service-user` part of this. @MikeMcQuaid you're likely best positioned to test that.

### DIFF
--- a/lib/service/formula_wrapper.rb
+++ b/lib/service/formula_wrapper.rb
@@ -122,12 +122,13 @@ module Service
 
     def owner
       if System.launchctl? && dest.exist?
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Looks good to me, thanks @Bo98!

PR #625, opened 2024-02-19
Improve behavior of `--all`

stefansundin:
This changes how `--all` behaves. When stopping all services, it filters out the services that are already stopped, and vice versa when starting all services.

My motivation for this change is that I have `dnscrypt-proxy` installed which causes `brew services stop --all` to no longer work properly and stop halfway through.

```
$ brew services list
Name           Status  User   File
caddy          none           
dnscrypt-proxy none    root   
grafana        started stefan ~/Library/LaunchAgents/homebrew.mxcl.grafana.plist

$ brew services stop --all
Warning: Service `caddy` is not started.
Error: Service `dnscrypt-proxy` is started as `root`. Try:
  sudo brew services stop dnscrypt-proxy

$ brew services list
Name           Status  User   File
caddy          none           
dnscrypt-proxy none    root   
grafana        started stefan ~/Library/LaunchAgents/homebrew.mxcl.grafana.plist
```

With my changes it works the way I want it to:

```
$ brew services list
Name           Status  User   File
caddy          none           
dnscrypt-proxy none    root   
grafana        started stefan ~/Library/LaunchAgents/homebrew.mxcl.grafana.plist

$ brew services stop --all
Stopping `grafana`... (might take a while)
==> Successfully stopped `grafana` (label: homebrew.mxcl.grafana)

$ brew services list
Name           Status User File
caddy          none        
dnscrypt-proxy none   root 
grafana        none        
```

You'll no longer see ```Warning: Service `X` is not started.``` warnings when using `--all`, which cleans up the output a lot.

If you have root services then `brew services start --all` no longer tries to start them. Perhaps a new switch should be included to allow this to be controlled, e.g. `--include-root`? In my opinion the default should exclude root services since I think it is the exception rather than the rule where you want to include them, right?

```
$ brew services start --all
==> Successfully started `caddy` (label: homebrew.mxcl.caddy)
==> Successfully started `grafana` (label: homebrew.mxcl.grafana)
```

The tests are still passing. Please let me know if you want me to make any changes. If you want me to add to the tests then please let me know. I don't want to spend time on tests unless the changes are desirable.

Thanks!

### DIFF
--- a/cmd/services.rb
+++ b/cmd/services.rb
@@ -102,12 +102,21 @@ module Homebrew
     opoo "The --all argument overrides provided formula argument!" if formula.present? && args.all?
 
     targets = if args.all?
-      ::Service::Formulae.available_services
+      if subcommand == "start"
+        ::Service::Formulae.available_services(loaded: false, skip_root: !::Service::System.root?)
+      elsif subcommand == "stop"
+        ::Service::Formulae.available_services(loaded: true, skip_root: !::Service::System.root?)
+      else
+        ::Service::Formulae.available_services
+      end
     elsif formula
       [::Service::FormulaWrapper.new(Formulary.factory(formula))]
     else
       []
     end
+
+    # Exit successfully if --all was used but there is nothing to do
+    return if args.all? && targets.empty?
 
     if ::Service::System.systemctl?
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense, great work @stefansundin!

PR #646, opened 2024-04-25
workflows: pin setup-ruby action

Moisan:
Pin the version of the `setup-ruby` action to full length commit SHA as described in the [security hardening for GitHub Actions guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).


MikeMcQuaid:
Thanks again @Moisan!

PR #711, opened 2024-08-17
Basic Sorbet `typed: true`

issyl0:
- I thought I'd use this as a test bed for https://github.com/Homebrew/brew/pull/18027. Now we have a minimal successful Sorbet typed tap.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -107,3 +107,15 @@ jobs:
       run: |
         rm -rf "$(brew --repo homebrew/services)"
         ln -s "$PWD" "$(brew --repo homebrew/services)"
+
+  sorbet:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Set up Homebrew
+      id: set-up-homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - run: brew typecheck homebrew/services
--- a/cmd/services.rbi
+++ b/cmd/services.rbi
@@ -1,0 +1,26 @@
+# typed: strict
+
+class Homebrew::Cmd::Services
+  sig { returns(Homebrew::Cmd::Services::Args) }
+  def args; end
+end
+
+class Homebrew::Cmd::Services::Args < Homebrew::CLI::Args
+  sig { returns(T::Boolean) }
+  def all?; end
+
+  sig { returns(T.nilable(String)) }
+  def file; end
+
+  sig { returns(T::Boolean) }
+  def json?; end
+
+  sig { returns(T::Boolean) }
+  def non_bundler_gems?; end
+
+  sig { returns(T::Boolean) }
+  def no_wait?; end
+
+  sig { returns(T.nilable(String)) }
+  def sudo_service_user; end
+end
--- a/lib/service.rbi
+++ b/lib/service.rbi
@@ -1,0 +1,5 @@
+# typed: strict
+
+module Service
+  include Kernel
+end
--- a/lib/service.rbi
+++ b/lib/service.rbi
@@ -2,4 +2,5 @@
 
 module Service
   include Kernel
+  include FileUtils
 end
--- a/lib/service/formula_wrapper.rb
+++ b/lib/service/formula_wrapper.rb
@@ -1,3 +1,4 @@
+# typed: true
 # frozen_string_literal: true
 
 # Wrapper for a formula to handle service-related stuff like parsing and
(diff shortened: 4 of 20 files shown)


cmd/services.rbi
```diff
@@ -0,0 +1,26 @@
+# typed: strict
+
+class Homebrew::Cmd::Services
+  sig { returns(Homebrew::Cmd::Services::Args) }
+  def args; end
+end
+
+class Homebrew::Cmd::Services::Args < Homebrew::CLI::Args
+  sig { returns(T::Boolean) }
```
issyl0, line 9: It would be cool to figure out generating these programmatically using the Tapioca compiler like we do with the main `brew` commands args, but also Copilot was helpful enough to do this manually for this MVP.

lib/service/commands/list.rbi
```diff
@@ -0,0 +1,5 @@
+# typed: strict
+
+module Service::Commands::List
+  include Kernel
+end
```
Bo98, line 5: If you do a single `service.rbi` that has

```rb
module Service
  include Kernel
end
```

it should apply to everything else.

We do something similar with the `Homebrew` module in Homebrew/brew.

lib/service/formula_wrapper.rb
```diff
@@ -10,9 +11,10 @@ class FormulaWrapper
     # Create a new `Service` instance from either a path or label.
     def self.from(path_or_label)
       return unless path_or_label =~ path_or_label_regex
+      return unless (formula = Regexp.last_match(1)).present?
 
       begin
-        new(Formulary.factory(Regexp.last_match(1)))
+        new(Formulary.factory(formula))
```
Bo98, line 17: Probably worth `T.must`ing this one so that it throws an error if it's missing (which should be programmer error as if the regex matches but is missing a group then the regex is wrong).

lib/service/services_cli.rb
```diff
@@ -89,6 +90,8 @@ def run(targets, verbose: false)
 
     # Start a service.
     def start(targets, service_file = nil, verbose: false)
+      file = T.let(nil, T.nilable(T.any(String, Pathname)))
```
Bo98, line 93:
```suggestion
      file = T.let(nil, T.nilable(Pathname))
```

I think?

Bo98 (approved):

MikeMcQuaid (commented):
Very nice!

lib/service/commands/list.rbi
issyl0: Hmm, apparently not, I tried that (`service.rbi` alongside `service.rb`) but it gave 38 errors all related to `Kernel` so I had to make individual RBIs?

issyl0:
I'll mark this as ready and re-run CI once https://github.com/Homebrew/brew/pull/18027 merges later today.

lib/service/commands/list.rbi
MikeMcQuaid: You can also do `include Kernel` in the actual modules.

Alternatively, I think in Homebrew/brew a `module.rbi` with the following might work:
```ruby
module Module
  include Kernel
end
```

Bo98: This is mostly caused by using `module_function` to be honest.

Should be able to remove all the RBIs when rebased on #714.

issyl0:
Thanks, @Bo98 - rebased this.

MikeMcQuaid (approved):
Much nicer, thanks @issyl0!

.github/workflows/tests.yml
```diff
@@ -107,3 +107,15 @@ jobs:
...
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Set up Homebrew
+      id: set-up-homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - run: brew typecheck homebrew/services
```
MikeMcQuaid, line 121: @issyl0 Think it'd be nice to put this into `brew test-bot --only-tap-syntax` for repos that have the necessary directories. Thoughts?

issyl0: Yes! I'll add that to my list.

lib/service.rbi
```diff
@@ -0,0 +1,6 @@
+# typed: strict
+
+module Service
+  include Kernel
+  include FileUtils
+end
```
Bo98, line 6: Do we need this file now?

Bo98 (approved):

lib/service.rbi
issyl0: You were correct. We do not.

PR #714, opened 2024-08-19
Drop unnecessary module_function

Bo98:
Should help #711 if CI passes

These modules should be called as `Service::Whatever.method` and never via `include Service::Whatever`, so remove `module_function` to forbid the latter and get stronger typing as a result.


issyl0 (approved):

MikeMcQuaid (approved):
This is great, thanks @Bo98!

PR #781, opened 2024-10-23
Improve handling of systemctl under mixed UID environments

Bo98:
When EUID != UID, `systemctl` basically ignores any environment variable and is largely nonfunctional.

`system_command` has handling to support this so wrap `systemctl` calls to make use of that.


SMillerDev (approved):

MikeMcQuaid (approved):
Makes sense, thanks @Bo98!

PR #800, opened 2024-11-06
Fix missing `max_wait` argument in `::Service::Commands::Stop.run` call

albertterradas:
### Summary
This PR fixes an issue in the `::Service::Commands::Stop.run` method call, where `max_wait` was referenced without being assigned a value.

### Changes
- Updated the `max_wait:` argument to `max_wait: max_wait` to correctly pass the intended `max_wait` variable.

### Context
The lack of a value assignment to `max_wait` resulted in an error when invoking the stop command. This change ensures that `max_wait` is now passed correctly as an argument.

### DIFF
--- a/cmd/services.rb
+++ b/cmd/services.rb
@@ -150,7 +150,7 @@ module Homebrew
           ::Service::Commands::Start.run(targets, args.file, verbose: args.verbose?)
         when *::Service::Commands::Stop::TRIGGERS
           max_wait = args.max_wait.to_f
-          ::Service::Commands::Stop.run(targets, verbose: args.verbose?, no_wait: args.no_wait?, max_wait:)
+          ::Service::Commands::Stop.run(targets, verbose: args.verbose?, no_wait: args.no_wait?, max_wait: max_wait)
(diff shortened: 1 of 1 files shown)


cmd/services.rb
```diff
@@ -150,7 +150,7 @@ def run
           ::Service::Commands::Start.run(targets, args.file, verbose: args.verbose?)
         when *::Service::Commands::Stop::TRIGGERS
           max_wait = args.max_wait.to_f
-          ::Service::Commands::Stop.run(targets, verbose: args.verbose?, no_wait: args.no_wait?, max_wait:)
+          ::Service::Commands::Stop.run(targets, verbose: args.verbose?, no_wait: args.no_wait?, max_wait: max_wait)
```
MikeMcQuaid, line 153: @albertterradas What Ruby version are you running? The existing syntax is valid on the version of Ruby that Homebrew requires.

MikeMcQuaid (changes requested):

p-linnane:
Closing due to lack of response.

PR #942, opened 2025-03-13
Deprecate tap

botantony:
Good to merge after Homebrew/brew#19385 is in a stable release.


Bo98:
Looks good, though we would need to wait until after the next brew release as those on the stable branch will only get the `brew` changes then (whereas updates to this tap are instant).

MikeMcQuaid (approved):

